PR #17, opened 2023-07-25
Merging DEVSKL from the old repo into the new repo

sklbancor:


sklbancor:
all tests pass at this stage; note that there were a number of API changes in the Optimizer module [I have split it into functional pieces] and if test coverage is not complete those may lead to issues when running the bot; therefore we should test it;

having said this -- all tests pass

mikewcasale (approved):

PR #91, opened 2023-08-27
69 add support for bancor v2 exchange

mikewcasale:


Lesigh-3100 (approved):
I've reviewed this by creating a Notebook, NB 050, that finds arbs that include a Bancor V2 pool, and goes through the process up to creating a route struct.

PR #96, opened 2023-08-30
61 implement replay mode

mikewcasale:


Lesigh-3100 (changes requested):
I noticed an issue that we are not getting Bancor V3 pool data. 

We might be able to solve this by adding:
(block_identifier=REPLAYBLOCK) to the bancor V3 contract call

Lesigh-3100 (dismissed):

PR #123, opened 2023-09-13
115 add support for automatic pool shutdown

Lesigh-3100:
Added a new arb mode, "pool_shutdown" that functions as a stand-alone mode to automatically shut down Bancor V3 pools.

While running this mode, the bot checks Bancor V3 Vault token balances vs staked balances. If the vault's token balance exceeds the staked balance, it will submit a transaction calling the withdrawPOL function for the token.


mikewcasale (changes requested):
Please review the changes I committed to `pool_shutdown.py` and `run_pool_shutdown.py`.

Lesigh-3100 (commented):
I looked over changes the & verified that it still works.

mikewcasale (approved):

PR #124, opened 2023-09-14
Optimization

sklbancor:
Making  a number of important changes to the core optimization algorithm to improve performance. Key changes are

- The **PairOptimizer** (previously SimpleOptimizer) is updated to ensure it is working correctly in the latest version; for single pairs this allows to do a marginal price optimization using a bisection algorithm which tends to converge better in some of the important edge cases we are experiencing

The below was part of the original spec but was moved to a future pull request. This one is reasonably risk free, and adding those convergence checks can break stuff if they are over zealous

- The **MargPOptimizer** is updated to allow for an additional convergence criterium (currently it uses "changes to the log prices are small"; we also add "the price is almost zero"*

*the issue with the second method is "almost zero" in crypto is not clear as it means very different things for BTC and for SHIB; generally the log changes (=percentage changes) is more financially robust; however -- in case of levered curves the target function can sit on a plateau (Jacobian = 0) in which case the gradient descent method fails


mikewcasale (dismissed):
In test 55, there is an import of class PairOptimizer which is not found. Please either fix the import statement or otherwise ensure the class in included in the codebase.

sklbancor:
> In test 55, there is an import of class PairOptimizer which is not found. Please either fix the import statement or otherwise ensure the class in included in the codebase.

can you recheck please? clearly this is not why the tests are failing [if there is ANY import error then this happens at test collection and none of the tests run at all, it does not even get there]

specifically the test that is failing is here
Screenshot 2023-09-16 at 14 19 05

and I can not replicate this on Jupyter; locally it works

NIXBNT (approved):

PR #202, opened 2023-11-14
Fix to use wei format for generating flashloans to avoid precision or rounding errors

Lesigh-3100:
-Using only 1 Flashloan token, taken from the first trade
-Use wei from trades to avoid rounding or precision errors


NIXBNT (approved):

zavelevsky (approved):
LGTM

PR #257, opened 2023-12-13
Click option for rpc

mikewcasale:
- Adds a new @click option to specify custom RPC urls
- Adds logic to override the config web3 connection with the specified RPC connection (if any)
- Adds a new @click option to disable written log-files


fastlane_bot/events/managers/base.py
```diff
@@ -321,6 +321,7 @@ def get_rows_to_update(self, update_from_contract_block: int) -> List[int]:
             for i, pool_info in enumerate(self.pool_data)
             if pool_info["last_updated_block"]
             < update_from_contract_block - self.alchemy_max_block_fetch
+            or str(pool_info["last_updated_block"]) == "nan"
```
barakman, line 324: This cast is dangerous.

Consider one of the following:
1. Converting that string to lowercase
2. Using `math.isnan` instead
3. Using `np.isnan()` instead
4. Using `pool_info["last_updated_block"] != pool_info["last_updated_block"]` instead

fastlane_bot/events/utils.py
```diff
@@ -560,6 +560,31 @@ def handle_flashloan_tokens(
...
+        WEB3_ALCHEMY_OPTIMISM = os.environ.get("WEB3_ALCHEMY_OPTIMISM")
+        rpc_url = f"https://opt-mainnet.g.alchemy.com/v2/{WEB3_ALCHEMY_OPTIMISM}"
+    elif rpc_url == "" and blockchain == "coinbase_base":
+        WEB3_ALCHEMY_BASE = os.environ.get("WEB3_ALCHEMY_BASE")
+        rpc_url = f"https://base-mainnet.g.alchemy.com/v2/{WEB3_ALCHEMY_BASE}"
+    elif rpc_url == "" and blockchain == "arbitrum_one":
+        WEB3_ALCHEMY_ARBITRUM = os.environ.get("WEB3_ALCHEMY_ARBITRUM")
+        rpc_url = f"https://arb-mainnet.g.alchemy.com/v2/{WEB3_ALCHEMY_ARBITRUM}"
+
+    return rpc_url
```
barakman, line 585: This function seems wrong in several different aspects:
1. Why does it need to take `rpc_url: str = None` as input to begin with?
2. Why does it check `not rpc_url` in some cases and `rpc_url == ""` in other cases?
3. Why does it check `rpc_url == ""` on every case? Surely there can be a simpler way
4. No final `else` case, is that why this function takes `rpc_url` as input to begin with?

It feels that there has to be a much less cumbersome way of writing this function, with the primary advantage being that it becomes less prone to unexpected behavior.

First off - no need to pass `rpc_url` just in order to return it as output if it contains a real value (i.e., neither `None` nor `""`).
Wherever this function is called, instead of passing that real `rpc_url` value, the caller can just use it instead of calling the function.
Subsequently everything becomes much cleaner in this function.
You can simply replace every `elif` statement with an `if` statement, and return the result directly from within each such statement.
And at the end of the function, in order to avoid unexpected behavior, you should raise an exception.

fastlane_bot/events/utils.py
```diff
@@ -598,13 +629,16 @@ def get_config(
 
     """
     default_min_profit_gas_token = Decimal(default_min_profit_gas_token)
+    if rpc_url:
+        rpc_url = get_rpc_url(blockchain, tenderly_fork_id, rpc_url)
```
barakman, line 633: Pointless, because at present, if `rpc_url` evaluates to a true result (i.e., it is a string of length larger than zero), then `get_rpc_url` simply returns this value as is. If anything, it should be called `if not rpc_url`.

fastlane_bot/events/utils.py
```diff
@@ -1310,6 +1350,39 @@ def get_latest_events(
...
+    Returns
+    -------
+    Tuple[int, int or None]
+        The starting block number and the block number to replay from.
+
+    """
+    if current_block is not None and start_block is not None:
+        if current_block - start_block > alchemy_max_block_fetch:
+            start_block = current_block - (alchemy_max_block_fetch + reorg_delay)
+    return start_block, replay_from_block
```
barakman, line 1383: Taking `replay_from_block` as input and returning it as output seems redundant in this function.

fastlane_bot/events/utils.py
```diff
@@ -1310,6 +1350,39 @@ def get_latest_events(
...
+
+    Returns
+    -------
+    Tuple[int, int or None]
+        The starting block number and the block number to replay from.
+
+    """
+    if current_block is not None and start_block is not None:
+        if current_block - start_block > alchemy_max_block_fetch:
+            start_block = current_block - (alchemy_max_block_fetch + reorg_delay)
```
barakman, line 1382: Semantic - I'd change this:
```
if current_block - start_block > alchemy_max_block_fetch:
    start_block = current_block - (alchemy_max_block_fetch + reorg_delay)
```
To this:
```
if start_block < current_block - alchemy_max_block_fetch:
    start_block = current_block - (alchemy_max_block_fetch + reorg_delay)
```
Because it makes it easier IMO to follow the logic behind the update of `start_block`.

barakman (changes requested):
How come lines such as:
```
self.WEB3_ALCHEMY_PROJECT_ID = network.WEB3_ALCHEMY_PROJECT_ID
self.RPC_URL = f"{network.RPC_ENDPOINT}{self.WEB3_ALCHEMY_PROJECT_ID}"
```
In file `provider.py`, haven't been modified in this PR?

PR #290, opened 2024-01-10
288 add read only setting

mikewcasale:


Lesigh-3100 (changes requested):
Needs a few bug fixes in async_event_update_utils - process_contract_chunks. Errors occur when running in read_only = True mode.

Lesigh-3100 (approved):
Problems fixed

PR #299, opened 2024-01-14
fix: local node gas parse error

joshmuente:
when hosting own node there is an bug in parsing message from the node.
I am running `geth version 1.13.9-stable-dd938d10`

` {'code': -32000, 'message': 'failed with 1679535 gas: max fee per gas less than block base fee: address 0xXXX, maxFeePerGas: 23816109243, baseFee: 24998394923'}`

`split2` is then parsed as `['22018912262,', "22998417652'}"]` with  `,` and `}"` chars that cant be parsed to an integer.

---
fix by simply removes the chars.


Lesigh-3100 (dismissed):

fastlane_bot/helpers/txhelpers.py
```diff
@@ -610,6 +610,7 @@ def build_transaction_with_gas(
                     message = str(e)
                     split1 = message.split("maxFeePerGas: ")[1]
                     split2 = split1.split(" baseFee: ")
+                    split2 = [split2[0], split2[1].replace("'}","")]
```
barakman, line 613: @joshmuente : can you please share the exact value of `message` (the variable in line 610 above)?

joshmuente: `{'code': -32000, 'message': 'failed with 1679535 gas: max fee per gas less than block base fee: address 0xXXX, maxFeePerGas: 23816109243, baseFee: 24998394923'}`

joshmuente: I also thought about first parsing the json via json.loads() and the splitting just the `message` in the json.
But I don't want to break support for alchemy thats why I just replaced the chars.

barakman: The variable `message` is set via:
```
message = str(e)
```
Which means that this variable is of type string.

Are you absolutely sure that this is the value of this variable is the string:
> "{'code': -32000, 'message': 'failed with 1679535 gas: max fee per gas less than block base fee: address 0xXXX, maxFeePerGas: 23816109243, baseFee: 24998394923'}"

And not the string:
> "failed with 1679535 gas: max fee per gas less than block base fee: address 0xXXX, maxFeePerGas: 23816109243, baseFee: 24998394923"

?

Thank you.

joshmuente: I have added logging to the variables like that: https://github.com/joshmuente/fastlane-bot/blob/a14450cca19a0743f682c9558ed6d8a13685a813/fastlane_bot/helpers/txhelpers.py#L611-L616

here is the output from the bot:
image

looks like `{'code': -32000, 'message': 'failed with 1646902 gas: max fee per gas less than block base fee: address 0xe8A9a2b134356CE433b9b4584bF66D9FB841d0E7, maxFeePerGas: 24288017269, baseFee: 24350587785'}`
is what I get

joshmuente: I found similar code here https://github.com/bancorprotocol/fastlane-bot/blob/d36287b8bca219f130e1bd1ad28a9f6811529139/fastlane_bot/tools/pool_shutdown.py#L264-L266

not sure if the same error applies to it

barakman:
Thank you for submitting the suggested fix.
PR #304 handles this in a more generic manner.
Thanks again!

PR #303, opened 2024-01-15
Log the type and value of each variable passed to function `get_start_block`

barakman:


Lesigh-3100 (changes requested):
This isn't something we want in general info-level logs. Can you please change it to debug?

barakman:
> This isn't something we want in general info-level logs. Can you please change it to debug?

If it gets masked out by some logging-level threshold or whatever, then it will miss out on the whole point in this PR to begin with, and we are left with the same logging output by the time the problem happens again.

mikewcasale (approved):

PR #304, opened 2024-01-15
Improve the handling of the error-message returned from the node

barakman:
Improve the handling of the error-message returned from the node to allow extracting the integer value correctly from any error-message which specifies it immediately after the substring `"baseFee: "`.
This change is designated for allowing interaction with different node types.


fastlane_bot/helpers/txhelpers.py
Lesigh-3100, line 1: @barakman could you please change the input on line 618 to: max_priority=max_priority

We don't want to change the max priority input as a reaction to the error message - only the Base fee. (This isn't your bug - preexisting mistake)

barakman: This PR does not intend to fix such bug (or any other bug for that matter).
It is designated for improving the extraction of the two integer values from the error-message received from the Alchemy node, in a manner which would make it compatible with other node types.

So it does not yield any **behavioral** change whatsoever.
Your suggested change **will**, so it should be done in a separate PR.

Lesigh-3100: Great, thank you - will make this a separate PR.

Lesigh-3100 (approved):

NIXBNT (approved):

PR #317, opened 2024-01-20
Move __init__ code to a new file bot.py

sklbancor:
# Issue

As previously discussed, importing anything from `fastlane_bot` takes a long time because when anything from the library is imported a lot of initialization code runs, including code that looks at the network. Also this code requires a lot of libraries that may or may not be present on all systems.

The `fastlane_bot` library at this stage is the canonical home of a number of modules / libraries that are important for analysis even without running the bot, and that can run on a stand-alone basis. Those libraries do not require the bot startup code to run, and they do not require any non-standard libraries to be installed on the system.

The issue that we are currently running into is **because the top __init__.py imports everybody and their dog from the fastlane_bot libraries, whenever you import anything from fastlane_bot, the whole shebang is run (or not, if packages are not present**.

# Solution

Move all offending imports for `__init__.py` to a new file called `bot.py`. Everyone using the bot has to slightly change the import statements (essentially `from fastlane_bot import xxx` becomes `from fastlane_bot.bot import xxx`, but this will be transparent for our end-users who use the script) but otherwise this is not a big change.


sklbancor:
out of date

PR #320, opened 2024-01-22
Fix the calculation of a non-integer block number

barakman:
If the `last_updated_block` column contains `None` values, then `max` returns a value of type `float`.
The caller should therefore verify that this value is nevertheless integer, and then convert it to type `int`.

Note that this PR also includes a technical refactoring of function `get_start_block`, aimed at simplifying the logical flow in this function, as well as the removal of log messages in this function (which were designated for the investigation of the non-integer-block-number issue).

### DIFF
--- a/fastlane_bot/events/utils.py
+++ b/fastlane_bot/events/utils.py
@@ -1335,49 +1335,33 @@ def get_start_block(
     Returns
     -------
     Tuple[int, int or None]
-        The starting block number.
+        The starting block number and the block number to replay from.
 
     """
-    if replay_from_block:
-        return (
-            replay_from_block - 1
-            if last_block != 0
-            else replay_from_block - reorg_delay - alchemy_max_block_fetch
-        ), replay_from_block
-    elif mgr.tenderly_fork_id:
-        # connect to the Tenderly fork and get the latest block number
-        from_block = mgr.w3_tenderly.eth.block_number
-        # Log all non-integer block numbers
-        non_int_values = [
-            (index, block["last_updated_block"], type(block["last_updated_block"]))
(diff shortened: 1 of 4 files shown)


mikewcasale (changes requested):
Couple things:

1. Let's add sub-functions (similar to `int_max()`) which handle each logical condition separately in a way that can be more easily unit-tested for each case, and include the corresponding unit tests for each case.
2. There needs to be handling for the case when the `current_block - last_updated_block > 2000`, in which case the correct return value should be current_block - 2000`. Without this handling, alchemy will throw errors and we cannot assume frequent events occur on obscure blockchains.

barakman:
> Couple things:
> 
> 1. Let's add sub-functions (similar to `int_max()`) which handle each logical condition separately in a way that can be more easily unit-tested for each case, and include the corresponding unit tests for each case.
> 2. There needs to be handling for the case when the `current_block - last_updated_block > 2000`, in which case the correct return value should be `current_block - 2000`. Without this handling, alchemy will throw errors and we cannot assume frequent events occur on obscure blockchains.

OK, but there are two issues regarding the 2nd bullet:

1. This is a logical change, which is outside the context of this PR, so it should probably be handled on a different PR.

2. The `2000` constant might be relevant for Alchemy but not for other providers (and in fact, even for `Alchemy` it may change occasionally as a result of new subscription-rules, or perhaps depending on the subscription-plan of each specific bot user, and so on). So it should not be forced here in this way, and perhaps it should not be forced here at all, but rather at a different point in the program, when blocks are actually fetched (for example, at the point where the program is about to fetch 8500 blocks, it can split that into 4 consecutive iterations of 2000 blocks each, followed by an additional iteration of 500 block).

mikewcasale (approved):

PR #322, opened 2024-01-22
Solidly curves

sklbancor:
This is the second part of the Solidly curve implementation (part one is [here](https://github.com/bancorprotocol/fastlane-bot/pull/315)). It adds a few further analytics, but more importantly, it now adds the ability to add Solidity curves to the optimizer.

This is implemented via the new CPC method `from_solidly` that should be called with the two token balances `x` and `y` (obviously consistent with the `pair` provided, in tokens as floats). 

**VERY IMPORTANTLY, THE METHOD RETURNS A LIST OF CURVERS, RATHER THAN A SINGLE CURVE**. The calling code should be able to handle the following situations

1. **Exactly one curve** is contained in the list; this is the current default case when the match works, ie the AMM is within the linear area and not in the wings
2. The **list is empty**, which means that the AMM is in the wings, and our flat approximation does not work, so we ignore it
3. There are **multiple curves** contained in the list; this case is not implemented yet, but it may be implemented soon. In this case, the traded against the Solidity curve should the _the sum of all trades against the returned curves__

## Integration instructions

The code in this PR is now ready for integration:
- get the (x,y,pair) values from the relevant solidly curves
- call CPC.from_solidly(...)
- this returns a list and the following cases should be dealt with (even though the last can't at the moment happen, but it may be easier to deal with it right away rather than having to change the code later again)
   - empty list -> the Solidly AMM is outside the range we approximate and should be ignored
   - one element -> the Solidly curve is approximated by a single CPC curve
   - multiple elements -> the Solidly curve is appriximated by multiple CPC curves
- when creating trade instruction, all dx and dy values across all curves returned should simply be added up* to obtain our best estimate of how the curve will react; note that this is an _approximation_ and it may be off by up to one percent at the current calibration and the code may have to deal with this

*as discussed previously, currently we only have one curve; when we have multiple curves you can be sure that the signs of the dx will all be the same (or zero), and ditto for the dy; most likely all but one will be maxed out, but you should not rely on it (and I don't think you care anyway)


fastlane_bot/tools/cpc.py
```diff
@@ -1555,6 +1700,46 @@ def isbigger(self, small, big):
             return big > self.EPS
         return big / small > 1 + self.EPS
```
barakman, line 1701: **Not directly related to this PR**.

The description below assumes `big, small >= 0` (otherwise this function might produce undesired output).

Assuming that the desired result in the general case here is to return check if the ratio between `big` and `small` is larger than one plus epsilon, the special case of `small == 0` should just return `big > 0`.

Note that it also works out mathematically:
- `big > 0 and small == 0` -->  `big / small  == inf > 1 + self.EPS`
- `big == 0 and small == 0` -->  `big / small tends to 1 < 1 + self.EPS`

That being said, I suspect the correct behavior to desire is slightly different, and would be achieved via:
```
return big > self.EPS + small
```
Again, assuming that both of these values are non-negative.

fastlane_bot/tools/cpc.py
```diff
@@ -1555,6 +1700,46 @@ def isbigger(self, small, big):
...
+        :func:                 function to plot (default: dyfrpmdx_f)
+        :show:                 if True, call plt.show()
+        :title:                plot title
+        :xlabel, ylabel:       axis labels
+        :grid:                 if True [False], [do not] show grid; None: ignore
+        :params:               additional kwargs passed to plt.plot
+        """
+        if xvals is None:
+            assert not xmin is None, "xmin must not be None if xv is None"
+            assert not xmax is None, "xmin must not be None if xv is None"
```
barakman, line 1718: Semantic - I'd change it to:
```
        if xvals == None:
            assert xmin != None, "xmin must not be None if xv is None"
            assert xmax != None, "xmin must not be None if xv is None"
```

fastlane_bot/tools/cpc.py
```diff
@@ -1555,6 +1700,46 @@ def isbigger(self, small, big):
...
+            x_v = np.linspace(xmin, xmax, steps) if steps else np.linspace(xmin, xmax)
+        else:
+            assert xmin is None, "xmin must be None if xv is not None"
+            assert xmax is None, "xmax must be None if xv is not None"
+            assert steps is None, "steps must be None if xv is not None"
```
barakman, line 1723: Semantic - I'd change `is` to `==`.

fastlane_bot/tools/cpc.py
```diff
@@ -1555,6 +1700,46 @@ def isbigger(self, small, big):
...
+            x_v = xvals
+            
+        xlabel = xlabel or (f"dx [{self.tknx}]" if not func else "x")
+        ylabel = ylabel or (f"dy [{self.tkny}]" if not func else "y")
```
barakman, line 1727: Using `or` between two values, one of which is a non-empty string.
Isn't that always going to produce `True`?

fastlane_bot/tools/invariants/functions/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+
+# objects defined in core
+from .core import Function, FunctionVector
+from .core import PriceFunction, Price2Function
+from .core import minimize, goalseek
+from .core import fmt
+
+# objects defined in funcs and funcsAMM
+from .funcs import *
+from .funcsAMM import *
```
barakman, line 32: I'm always in favor of importing exactly what's needed (for readability, not other impact as far as I'm aware of).

fastlane_bot/tools/invariants/functions/funcs.py
```diff
@@ -0,0 +1,81 @@
...
+
+(c) Copyright Bprotocol foundation 2024. 
+Licensed under MIT
+"""
+#__VERSION__    ==> core.__VERSION__
+#__DATE__       ==> core.__DATE__
+
+from .core import Function as _Function, dataclass as _dataclass
+import math as _m
+
```
barakman, line 12: You might wanna use, for example, `a: float = 0.0` throughout this file.
Otherwise the declared variable is of type `int` regardless of the `float` hint.

fastlane_bot/tools/invariants/vector.py
```diff
@@ -101,6 +101,22 @@ def _kwargs(self, other=None):
         """
         return dict()   
     
+    @property
+    def elements(self):
+        """returns the elements (keys!) of the vector as a list"""
+        return list(self.vec.keys())
+    el = elements
```
barakman, line 108: This global scope line looks out of place (did you really intend for it to be here?).

fastlane_bot/tools/cpc.py
sklbancor: > the special case of small == 0 should just return big > 0.

not quite sure I am using it at all, but the idea here is that for small == big you are guaranteed False

sklbancor: not sure the case ``small == 0`` is particularly relevant, but I did want to handle it
thing is -- normally I want EPS to be a dimensionless constant because ``big > small + self.EPS`` means different things then ``small = 1e-20`` vs ``small = 1e20``. That is the case that I want to handle well -- especially in crypto where we can have prices in both orders of magnitude.

The case `small=0` is an afterthought and given that EPS will the 1e-3 ... 1e-9 it seems good enough

fastlane_bot/tools/invariants/functions/__init__.py
sklbancor: I generally agree but those files are designed to _only_ contain example functions, and I simply want to import all of them without having to adjust the import statement whenever I add examples

fastlane_bot/tools/cpc.py
barakman: 1. I cannot see the equivalence here
2. I cannot see how it's related to what I wrote

fastlane_bot/tools/invariants/vector.py
sklbancor: yeah; `el` is simply an alias for `elements`, typically for interactive use; I do that quite often

fastlane_bot/tools/cpc.py
sklbancor: the expression works as follows

xlabel given -> xlabel
not func -> f"dx [{self.tknx}]" 
else -> x

sklbancor: Screenshot 2024-03-21 at 13 17 03

barakman: I am not familiar with `or` being anything other than a boolean operator (i.e., the expression `x or y` is always evaluated to either True or False as far as I'm aware of).

And even if this is not the case, then for the purpose of readability, I'd restrict my usage of this operator only as a boolean one.

As far as I understand `xlabel = xlabel or (f"dx [{self.tknx}]" if not func else "x")`, it is equivalent to:
```
if not xlabel:
    xlabel = "x" if func else f"dx [{self.tknx}]"
```
If this analysis is correct, then I'd rewrite that statement as such (again, for the purpose of readability).

sklbancor: Barak, respectfully: we are writing code for people who are familiar with Python; if someone stumbles upon this expression which uses a well-known Python construct and wonders how it works then they should probably take some time to look up the finer details of the language and not assume it is JavaScript or C++

barakman: You asked for a code review, I'm giving you one; feel free to do whatever you want with it :)

zavelevsky: siding with @sklbancor here - as it's pretty standard for OR expression to serve as means to return the first truthy value. Even in JS for example:
image

zavelevsky: But this is a param passed from outside - unless you're suggesting using this option in the default params?

sklbancor: regarding efficiency -- keep in mind that everything in Python is an object, so essentially that is a simple 64-bit write, or whatever the length of an address field is. In an interpreted language this does not take any meaningful time, relatively speaking.

barakman: BTW, you may by the least want to change your "if not func" logic to "if func", which - again - makes it slightly easier to read. In either case, this thread has spanned across 998 more comments than what it was initially designated for, so feel free to resolve the comment and fix the code whichever way you see fits.

sklbancor: Fun fact:

`x = x or None`  is actually (slightly) faster (!!!) than the version with if

Screenshot 2024-03-21 at 14 35 22

barakman (approved):

PR #324, opened 2024-01-23
hex string without 0x prefix

mikewcasale:
…erly handled

- Carbon cid values are provided by the contract and are int types, whereas all other exchange cid values are generated by hashing the descr (exchange tkn1/tkn2 fee) info.
- This fix ensures that the types and values of carbon cid's are preserved
- This change is also believed to be a fix for the seemingly unrelated `Ox string without prefix` error

### DIFF
--- a/fastlane_bot/events/async_event_update_utils.py
+++ b/fastlane_bot/events/async_event_update_utils.py
@@ -474,7 +474,7 @@ def async_update_pools_from_contracts(mgr: Any, current_block: int, logging_path
     # Initialize web3
     new_pool_data_df["cid"] = [
         cfg.w3.keccak(text=f"{row['descr']}").hex()
-        if row["exchange_name"] != "carbon_v1"
+        if row["exchange_name"] not in mgr.cfg.CARBON_V1_FORKS
         else int(row['cid'])
         for index, row in new_pool_data_df.iterrows()
     ]
--- a/fastlane_bot/events/async_event_update_utils.py
+++ b/fastlane_bot/events/async_event_update_utils.py
@@ -474,6 +474,8 @@ def async_update_pools_from_contracts(mgr: Any, current_block: int, logging_path
     # Initialize web3
     new_pool_data_df["cid"] = [
         cfg.w3.keccak(text=f"{row['descr']}").hex()
+        if row["exchange_name"] != "carbon_v1"
+        else int(row['cid'])
         for index, row in new_pool_data_df.iterrows()
     ]
 


fastlane_bot/events/async_event_update_utils.py
```diff
@@ -474,6 +474,8 @@ def async_update_pools_from_contracts(mgr: Any, current_block: int, logging_path
     # Initialize web3
     new_pool_data_df["cid"] = [
         cfg.w3.keccak(text=f"{row['descr']}").hex()
+        if row["exchange_name"] != "carbon_v1"
+        else int(row['cid'])
```
barakman, line 478: First part (the `if`) returns a string (`str`), second part (the `else`) returns an integer (`int`).

This is not ideal IMO.

Can you please make sure that both parts return a result of the same type AND of the same format?

For example, if you return a string to the caller function, and that function needs to turn it into an integer, then it will to use:
- `int(retVal)` for a string which represents an integer in decimal base
- `int(retVal, 16)` for a string which represents an integer in hexadecimal base

Which more generally means that a value of an inconsistent type OR an inconsistent returned from the callee, would propagate that inconsistency to the caller.

Presumably, if the caller function has been handling a hexadecimal string up until now, then in order to avoid changing that function, you should stick to that same type and that same format.

You can do that by replacing this:
```
int(row['cid'])
```
With this:
```
hex(int(row['cid']))
```
If you determine that returning a plain integer (`int`) in BOTH cases is a better choice, then you'll most likely need to adjust the caller function accordingly.

barakman (changes requested):

fastlane_bot/events/async_event_update_utils.py
mikewcasale: While I can understand the desire and reasoning of the request, there is no simple way to fulfill it. Currently, the `cid` value is the unique identifier for all pools tracked in the system, and ONLY Carbon is an integer type. In fact, if you return a hex(str(int)) value for Carbon pools here, you will end up with the same error that this change attempts to fix. Moreover, if you return a str(int) type for Carbon pools, it will also fail downstream and/or produce unexpected results where there are assumptions that Carbon pools will be of type int. 

Thus, the only way to complete your request would be to re-architect the primary-key value (`cid`) of the pool data lookup table, such that it contain all of one or the other types. However, I suggest we attempt something of this sort in a dedicated PR, as it will likely require a bunch of changes throughout the system.

PR #327, opened 2024-01-23
293 reorganize testing infrastructure

mikewcasale:
To run the new test infrastructure, use `pytest tests`.


tests/test_002_CPCandOptimizer.py
```diff
@@ -26,10 +26,7 @@
 
 #
 
-try:
-    market_df = pd.read_csv("_data/NBTEST_002_Curves.csv.gz")
-except:
-    market_df = pd.read_csv("fastlane_bot/tests/nbtest/_data/NBTEST_002_Curves.csv.gz")
+market_df = pd.read_csv("tests/_data/NBTEST_002_Curves.csv.gz")
```
barakman, line 29: Better use general utility function `read_csv_file` (in file `utils.py`).
The reason being, it would make it easy to apply a change to all similar calls, in a single place.

tests/test_002_CPCandOptimizer.py
```diff
@@ -1211,7 +1208,7 @@ def test_simple_optimizer():
     funcvy = O.optimize(result=O.SO_DXDYVALYFUNC)
     x,y = func0(2100)[0]
     xb, yb, _ = c0.dxdyfromp_f(2100)
-    assert x == xb
+    assert x == xb, f"x={x}, xb={xb}"
     assert y == yb
```
barakman, line 1212: Consider adding an assertion-message here too.
Also, consider making both of these messages more verbose (which will also make it easier to find each them in the code).

tests/test_002_CPCandOptimizer.py
```diff
@@ -1472,7 +1469,7 @@ def test_margp_optimizer():
     assert set(r.tokens_t) == {'USDC', 'USDT'}
     assert r.errormsg is None
     assert r.is_error == False
-    assert r.time > 0
+    assert r.time >= 0
```
barakman, line 1472: This looks like a functional (behavioral) change; Are you sure that this is the right thing to do?
Even if yes, are you sure that it should be done in this PR?

tests/test_002_CPCandOptimizer.py
```diff
@@ -1500,7 +1497,7 @@ def test_margp_optimizer():
     assert set(r.tokens_t) == set(('USDC', 'USDT'))
     assert r.errormsg is None
     assert r.is_error == False
-    assert r.time > 0
+    assert r.time >= 0
```
barakman, line 1500: Same as above

tests/test_002_CPCandOptimizer.py
```diff
@@ -1548,7 +1545,7 @@ def test_margp_optimizer():
...
```
barakman, line 1548: Same as above

tests/test_004_GraphCode.py
```diff
@@ -622,10 +626,7 @@ def test_interaction_with_cpc():
 def test_with_real_data_from_cpc():
 # ------------------------------------------------------------
     
-    try:
-        df = pd.read_csv("_data/NBTEST_002_Curves.csv.gz")
-    except:
-        df = pd.read_csv("fastlane_bot/tests/nbtest/_data/NBTEST_002_Curves.csv.gz")
+    df = pd.read_csv("tests/_data/NBTEST_002_Curves.csv.gz")
```
barakman, line 629: Better use general utility function `read_csv_file` (in file `utils.py`)...

barakman (changes requested):

tests/test_002_CPCandOptimizer.py
barakman: Sorry, that's a test file, so probably no issue (leaving the previous comment in place, just in case it IS an issue).

tests/test_004_GraphCode.py
barakman: Perhaps not critical in test files (i.e., only in core module files).

resources/analysis/202401 Solidly/202401 Solidly-Freeze01.ipynb
sklbancor, line 1: You cannot delete / add gitignored file types; if you need to move them somewhere else you need to use git move (or git add -f). On a different note -- the purpose of this PR is to reorganise the testing infrastructure so I would strongly suggest it does only that and does not touch anything else.

In other words -- any commit touching files that are not test related should be removed

sklbancor (changes requested):
1. This PR touches areas that are unrelated to tests. All those should be removed. 

1. This PR also hardcodes current tests that relate to the tools module. What should happen there is that
1/ those tests are moved as jupyter notebooks into their target destination (in tools, or outside of it, Doron's choice)
2/ the conversion scripts are adapted (the generated tests should NOT end up in the same folder as the hardcoded ones)
3/ the test script is adapted to include the converted tests

1. this PR moves tests from the 900 area (slow tests) into the common area; this should not happen: slow tests must run last

1. run_tests is broken (see below); also tests are apparently no longer run automatically via github actions; github actions should run at least the fast tests (those with number <900), including those in tools, before a pull request can be merged

resources/analysis/202401 Solidly/202401 Solidly-Freeze01.ipynb
sklbancor: This happens with run_tests

```
(base) skl@MBA21 FLBot % scripts/run_tests 
/Users/skl/REPOES/Bancor/FLBot/scripts
mkdir: tests: No such file or directory
mkdir: tests/nbtest: No such file or directory
cp: directory tests/nbtest/_data does not exist
touch: tests/__init__.py: No such file or directory
touch: tests/nbtest/__init__.py: No such file or directory
python: can't open file '/Users/skl/REPOES/Bancor/FLBot/scripts/fastlane_bot/tools/resources/NBTest/ConvertNBTest.py': [Errno 2] No such file or directory
============================= test session starts ==============================
platform darwin -- Python 3.11.5, pytest-7.4.0, pluggy-1.0.0 -- /Users/skl/anaconda3/bin/python
cachedir: .pytest_cache
rootdir: /Users/skl/REPOES/Bancor/FLBot
plugins: anyio-3.5.0
collected 0 items                                                              

============================ no tests ran in 0.00s =============================
ERROR: file or directory not found: tests
```

sklbancor: Screenshot 2024-02-06 at 13 03 00

PR #328, opened 2024-01-23
326 fix production job alert spam time outs

mikewcasale:
This should have no effect on system usage/users, and will only be noticed in the production environment.


barakman (commented):
Nicely done!
Why is file `run_pool_shutdown.py` removed?

mikewcasale:
> Nicely done! Why is file `run_pool_shutdown.py` removed?

It is somewhat unrelated, so I will add the file removal to a new PR

main.py
```diff
@@ -62,331 +61,290 @@
...
+def process_arguments(args):
+    """
+    Process and transform command line arguments.
+
+    :param args: Namespace object containing command line arguments.
+    :return: Processed arguments.
+    """
+    # Define the transformations for each argument
+    transformations = {
+        "backdate_pools": lambda x: x == "True",
```
barakman, line 78: Is this the standard way of doing it?
I assume that you want a configuration parameter which can be either mentioned or omitted, but with no additional input following that, so it serves strictly as a boolean flag.
Wouldn't it be more correct to use something like:
```
lambda x: True if x else False
```
Or even:
```
lambda x: bool(x)
```
Same goes for a few other similar ones below...?

barakman (changes requested):

zavelevsky (approved):
LGTM

main.py
mikewcasale: Your suggestions do not handle the context properly. The command-line arguments are received as strings. I have attached three example usages, the first two are your suggestions, and the third is the existing code. You'll notice that only the third result gives the correct output. 
image

barakman: That's not how I meant for these boolean flags to be used.

What I meant is if you specify `--someBooleanOption` (without any additional value ascribed to it), then it means that you want that option enabled. And if you do not specify`--someBooleanOption`, then it means that you do not want this option enabled.

For this type of usage, I believe that `lambda x: True if x else False` should do the job.

mikewcasale: Please test your proposed changes prior to suggesting their use. Argparse has certain behaviors which are out of our control and out of scope for this PR. I am not familiar with any way to achieve what you're suggesting here, so please propose the change you'd like to see, if indeed it is possible.

barakman: Something like this, for example:
```
from argparse import ArgumentParser
from argparse import BooleanOptionalAction

parser = ArgumentParser()
parser.add_argument('--boolean_option', action=BooleanOptionalAction)

args = parser.parse_args()
print(args.boolean_option)
```
It returns `True` when `--boolean_option` is specified, and `None` when `--boolean_option` is omitted.

If the type-duality matters, then you can of course convert it to `bool`, which will give you `True` and `False`.

mikewcasale: OK this is now updated in a way similar to what you proposed.

barakman (approved):

barakman (approved):

PR #332, opened 2024-01-25
Add function native_gas_token_to_wrapped and remove Ethereum only forconversion in routehandler.py _calc_carbon_output

Lesigh-3100:
Bug Description
For strategies that include native ETH on Graphene, there is an assertion error in the calculated trade instructions because of a line that only converts NATIVE > WRAPPED gas token for Ethereum.

Severity (High/Medium/Low)
High

Steps to Reproduce
Search for strategies with native ETH on Graphene.

Expected Behavior
The bot should not throw an assertion error. It should treat native gas token as wrapped gas token for the trade calculation.

Actual Behavior
An assertion error is thrown because the bot is not converting the address.

Impact Analysis (to be filled by contributors)
Automated Tests: Are there existing tests covering this? Do we need new tests?
A test should be added that checks that native gas token is being converted to wrapped gas token for trade calculations.

Risks: Potential areas affected by this bug and its probable fix
My concern is mixing trades between wrapped & native gas token strategies. This can be addressed by confirming that a separate trade would be generated for a WRAPPED/TKN strategy & a NATIVE/TKN strategy when trading on both.

Performance: Is this bug affecting performance?
No

Monitoring: Does this affect any current monitoring? Any new alerts needed?
No

Logging: Any changes required in logging?
No

Documentation: Updates needed in documentation?
No


fastlane_bot/helpers/routehandler.py
```diff
@@ -1358,14 +1375,13 @@ def _calc_carbon_output(
         Decimal
             The amount out.
         """
+        assert tkn_in not in self.ConfigObj.NATIVE_GAS_TOKEN_ADDRESS,"[routehandler.py _calc_carbon_output] This error is appearing because the function does not expect an input of the native gas token. Please post this error with your run configuration in the Bancor Developers Telegram channel."
```
barakman, line 1378: You should probably make the log entry much shorter, as log entries only need to:
1. To include a high-level description of the problem, plus the related state (variables) if relevant
2. To be easily (i.e., uniquely) traceable in the code

You can probably add most of that description as a comment in the code.

fastlane_bot/helpers/routehandler.py
```diff
@@ -555,6 +555,23 @@ def _get_flashloan_struct(self, trade_instructions_objects: List[TradeInstructio
 
         return flashloans
 
+    @staticmethod
+    def native_gas_token_to_wrapped(tkn: str, wrapped_address: str, native_address: str):
```
barakman, line 559: You can make this a member method instead of a static method, i.e., take `self` as the input argument, and then:
- Instead of passing `wrapped_address` as input, you can obtain it directly from `self.ConfigObj.WRAPPED_GAS_TOKEN_ADDRESS`
- Instead of passing `native_address` as input, you can obtain it directly from `self.ConfigObj.NATIVE_GAS_TOKEN_ADDRESS`

It will also make your calls to this function (further below) much cleaner and easier to read.

resources/NBTest/NBTest_062_TestRouteHandler.ipynb
```diff
@@ -10,7 +10,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "imported m, np, pd, plt, os, sys, decimal; defined iseq, raises, require\n",
+      "imported m, np, pd, plt, os, sys, decimal; defined iseq, raises, require, Timer\n",
```
barakman, line 13: How is this change related to the problem that this PR is attempting to fix?

barakman (changes requested):

fastlane_bot/helpers/routehandler.py
Lesigh-3100: I considered this but decided to make it static to be able to remove its dependency on the Config object.

resources/NBTest/NBTest_062_TestRouteHandler.ipynb
Lesigh-3100: I did not manually change this specifically. I added a test to this Notebook to test the function that I added - native_gas_token_to_wrapped

fastlane_bot/helpers/routehandler.py
Lesigh-3100: I shortened the error message.

fastlane_bot/helpers/routehandler.py
barakman: This argument doesn't make any sense whatsoever.
If you ever change or remove the Config-Object dependency in this class, then you do it in this function, and not in every other function which calls it.
As it stands, this function does nothing other than `return x if condition else y`.
It is not specific to the ETH/WETH issue other than by the names that you chose for the function itself, as well as to each one of its parameters.
In that case, you may as well move it to the general-utility module, and name it something generic...
Or alternative, simply do the `return x if condition else y` part directly where you call this function (which pretty much means reverting this change altogether, if I remember correctly).
A third option would be to pass the `self.Config` object to this method, making it indeed specific to the ETH/WETH issue (since it will then read the related attributes from that object).
But again - if you choose to implement this method **in this class**, then the most correct way to do so is by letting it handle everything (meaning - it takes `self` as input, and performs everything else internally).

resources/NBTest/NBTest_062_TestRouteHandler.ipynb
barakman: Well, where is that test then?
Something doesn't make sense here.

fastlane_bot/helpers/routehandler.py
Lesigh-3100: Ok, that makes sense. I implemented your suggestion of removing static from the function & using the Config object.

fastlane_bot/helpers/routehandler.py
barakman: If the stuff that you removed is important, then please add it as a comment above the assertion line.

fastlane_bot/helpers/routehandler.py
barakman: Have you fixed accordingly, every place you call this function?

resources/NBTest/NBTest_062_TestRouteHandler.ipynb
Lesigh-3100: NBTest_062_TestRouteHandler.py

barakman (changes requested):

fastlane_bot/helpers/routehandler.py
```diff
@@ -555,6 +555,17 @@ def _get_flashloan_struct(self, trade_instructions_objects: List[TradeInstructio
...
+    def native_gas_token_to_wrapped(self, tkn: str):
+        """
+        This function returns the wrapped gas token address if the input token is the native gas token, otherwise it returns the input token address.
+        :param tkn: str
+            The token address
+        returns: str
+            The token address, converted to wrapped gas token if the input was the native gas token
+
+        """
+        return self.ConfigObj.WRAPPED_GAS_TOKEN_ADDRESS if tkn in self.ConfigObj.NATIVE_GAS_TOKEN_ADDRESS else tkn
```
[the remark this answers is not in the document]
Lesigh-3100: Sure - I changed this to ==

fastlane_bot/helpers/routehandler.py
Lesigh-3100: It's not important it was just verbose.

fastlane_bot/helpers/routehandler.py
```diff
@@ -1358,14 +1369,13 @@ def _calc_carbon_output(
         Decimal
             The amount out.
         """
+        assert tkn_in not in self.ConfigObj.NATIVE_GAS_TOKEN_ADDRESS,"[routehandler.py _calc_carbon_output] Function does not expect native gas token as input."
```
barakman, line 1372: Let's start get rid of these `in` and `not in`, shall we?

Using `!=` here would do the job fine AND be more readable (and probably also faster).

barakman (changes requested):

fastlane_bot/helpers/routehandler.py
Lesigh-3100: Changed to !=

mikewcasale (approved):

Issue #331, opened 2024-01-25
Trading on aerodrome_v2 is not properly supported


barakman:
## Bug Description
- When an arbitrage route includes `aerodrome_v2` the bot throws an exception and reiterate the whole process (i.e., the exception is caught only at the "main loop" layer).

## Severity (High/Medium/Low)
- High

## Steps to Reproduce
- Simulate an arbitrage opportunity that goes through `aerodrome_v2`

## Expected Behavior
- Should complete successfully

## Actual Behavior
- Throws an exception

## Detailed Description

The exception most likely occurs in file `solidly_vs.py`:
```
async def get_fee(self, address: str, contract: Contract) -> Tuple[str, float]:
    if "velocimeter" in self.exchange_name:
        default_fee = await self.factory_contract.caller.getFee(address)
```
With `self.factory_contract` not being initialized (i.e., set to `None`).

Which is most likely the result of the following (in file `base.py`):
```
def handle_solidly_exchanges(self, exchange):
    """
    Handles getting stable & volatile fees for Solidly forks
    """
    exchange_name = exchange.exchange_name
    self.factory_contracts[exchange_name] = self.web3.eth.contract(
        address=self.cfg.FACTORY_MAPPING[exchange_name],
        abi=exchange.get_factory_abi,
    )
    exchange.factory_contract = self.factory_contracts[exchange.exchange_name]
```
With `exchange.exchange_name` not initialized correctly (set to `"solidly_v2"` instead of `"velodrome_v2"`).

Which is most likely the result of the following (in file `base.py`):
```
if base_exchange_name in "solidly_v2":
    self.exchanges[exchange_name] = self.handle_solidly_exchanges(exchange=self.exchanges[exchange_name])
```
With `exchange_name` not initialized correctly (set to `"solidly_v2"` instead of `"velodrome_v2"`).

PR #333, opened 2024-01-25
331 trading on aerodrome v2 is not properly supported

Lesigh-3100:
This PR addresses the issue:
https://github.com/bancorprotocol/fastlane-bot/issues/331

The bug occurred in async_event_update_utils.py specifically for Solidly-fork exchanges. 

The issue was that Solidly forks require the Factory contract to get the pool's fee, and when get_fee was called in the function get_token_and_fee, the exchange object had not been properly initialized with its factory contract. I changed the function to use the exchange object that was already initialized on startup, stored in Manager base.py.

### DIFF
--- a/fastlane_bot/events/async_event_update_utils.py
+++ b/fastlane_bot/events/async_event_update_utils.py
@@ -361,7 +361,7 @@ def get_pool_contracts(mgr: Any) -> List[Dict[str, Any]]:
     contracts = []
     for add, en, event, key, value in mgr.pools_to_add_from_contracts:
         exchange_name = mgr.exchange_name_from_event(event)
-        ex = exchange_factory.get_exchange(key=exchange_name, cfg=mgr.cfg, exchange_initialized=False)
+        ex = mgr.exchanges[exchange_name]
         abi = ex.get_abi()
         address = event["address"]
         contracts.append(
--- a/fastlane_bot/events/async_event_update_utils.py
+++ b/fastlane_bot/events/async_event_update_utils.py
@@ -2,8 +2,10 @@ import asyncio
 import os
 import time
 from glob import glob
+from random import shuffle
 from typing import Any, List, Dict, Tuple, Type, Callable
 
+import nest_asyncio
 import numpy as np
 import pandas as pd
 from pandas import DataFrame
@@ -12,10 +14,11 @@ from web3.contract import AsyncContract
 
 from fastlane_bot.data.abi import ERC20_ABI
 from fastlane_bot.events.async_utils import get_contract_chunks
-from fastlane_bot.events.exchanges import exchange_factory
+from fastlane_bot.events.exceptions import AyncUpdateRetryException
 from fastlane_bot.events.utils import update_pools_from_events
-import nest_asyncio
+
 nest_asyncio.apply()
+
 
 async def get_missing_tkn(contract: AsyncContract, tkn: str) -> pd.DataFrame:
     try:
@@ -58,7 +61,7 @@ async def main_get_missing_tkn(c: List[Dict[str, Any]]) -> pd.DataFrame:
 
 
 async def get_token_and_fee(
-    exchange_name: str, ex: Any, address: str, contract: AsyncContract, event: Any
+        exchange_name: str, ex: Any, address: str, contract: AsyncContract, event: Any
 ) -> Tuple[str, str, str, str, str, int or None, str or None] or Tuple[
     str, str, None, None, None, None, None
 ]:
@@ -132,22 +135,22 @@ async def main_get_tokens_and_fee(c: List[Dict[str, Any]]) -> pd.DataFrame:
 
 
 def pair_name(
-    t0_symbol: str,
-    tkn0_address: str,
-    t1_symbol: str,
-    tkn1_address: str,
-    key_digits: int = 4,
+        t0_symbol: str,
+        tkn0_address: str,
+        t1_symbol: str,
+        tkn1_address: str,
+        key_digits: int = 4,
 ) -> str:
     return f"{t0_symbol}-{tkn0_address[-key_digits:]}/{t1_symbol}-{tkn1_address[-key_digits:]}"
 
 
 def get_pool_info(
-    pool: pd.Series,
-    mgr: Any,
-    current_block: int,
-    tkn0: Dict[str, Any],
-    tkn1: Dict[str, Any],
-    pool_data_keys: frozenset,
+        pool: pd.Series,
+        mgr: Any,
+        current_block: int,
+        tkn0: Dict[str, Any],
+        tkn1: Dict[str, Any],
+        pool_data_keys: frozenset,
 ) -> Dict[str, Any]:
     fee_raw = eval(str(pool["fee"]))
     pool_info = {
@@ -185,6 +188,7 @@ def get_pool_info(
 
     return pool_info
 
+
 def sanitize_token_symbol(token_symbol: str, token_address: str, read_only: bool) -> str:
     """
     This function ensures token symbols are compatible with the bot's data structures.
@@ -207,7 +211,7 @@ def sanitize_token_symbol(token_symbol: str, token_address: str, read_only: bool
 
 
 def add_token_info(
-    pool_info: Dict[str, Any], tkn0: Dict[str, Any], tkn1: Dict[str, Any], read_only: bool
+        pool_info: Dict[str, Any], tkn0: Dict[str, Any], tkn1: Dict[str, Any], read_only: bool
 ) -> Dict[str, Any]:
     print(f"called add_token_info")
     tkn0_symbol = tkn0["symbol"].replace("/", "_").replace("-", "_")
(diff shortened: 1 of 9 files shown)


fastlane_bot/events/async_event_update_utils.py
```diff
@@ -361,13 +361,13 @@ def get_pool_contracts(mgr: Any) -> List[Dict[str, Any]]:
     contracts = []
     for add, en, event, key, value in mgr.pools_to_add_from_contracts:
         exchange_name = mgr.exchange_name_from_event(event)
-        ex = exchange_factory.get_exchange(key=exchange_name, cfg=mgr.cfg, exchange_initialized=False)
+        ex = mgr.exchanges[exchange_name]
```
barakman, line 364: The code which you've removed:
```
exchange_factory.get_exchange(key=exchange_name, cfg=mgr.cfg, exchange_initialized=False)
```
Is used in a few other places.

Why was it used like this here, and in each one of the other places?

Are you sure that changing it only here is the correct thing to do?

The more general question to ask here - why is the `exchange_factory` object not just a simple dictionary? Finding the exact reason for this will probably answer all the other questions above.

Lesigh-3100: Thank you! I updated this in the other places that I found it too. 

Now exchange_factory.get_exchange is only used once on startup, and other times it is referenced from the dictionary stored in the base manager.

fastlane_bot/events/async_utils.py
```diff
@@ -12,8 +12,6 @@ def get_contract_chunks(contracts: List[Dict[str, Any]]) -> List[List[Dict[str,
 
 def get_abis_and_exchanges(mgr: Any) -> Dict[str, Any]:
     abis = {}
-    exchanges = {}
     for exchange in mgr.exchanges:
-        exchanges[exchange] = exchange_factory.get_exchange(key=exchange, cfg=mgr.cfg, exchange_initialized=False)
-        abis[exchange] = exchanges[exchange].get_abi()
+        abis[exchange] = mgr.exchanges[exchange].get_abi()
```
barakman, line 16: This code looks extremely suspicious:
```
for exchange in mgr.exchanges:
    abis[exchange] = mgr.exchanges[exchange].get_abi()
```
First thing that comes to mind is that you may as well do:
```
for exchange in mgr.exchanges:
    abis[exchange] = exchange.get_abi()
```
But then the question that arises is - what exactly is `exchange`?
Because on the one hand, it is used as a key to the `abis` dictionary.
And on the other hand, it is used as an object calling one of its class methods (`get_abi`).

Python allows using objects as keys as far as I remember, but is this indeed what we're doing?
Have you verified that it works as desired?

fastlane_bot/events/async_event_update_utils.py
barakman: I've found 4 places in the code where we call function `get_exchange`, and pass to it more than just the exchange name:
- 2 in file async_event_update_utils.py
- 1 in file async_utils.py
- 1 in file base.py

I'm assuming that the one in file base.py is the one that your refer to as "once on startup".

Have you addressed all the others?

Lesigh-3100: Yes - I changed this in each location except for base.py.

fastlane_bot/events/async_utils.py
mikewcasale: Your comment that you might as well do the following:

```python
for exchange in mgr.exchanges:
    abis[exchange] = exchange.get_abi()
```

 is not correct. The `mgr.exchanges` object is structured as follows, where `ExchangeA, ExchangeB...` are instances of the `Exchange` class.

```python    
mgr.exchanges = {"exchange_a": ExchangeA, "exchange_b": ExchangeB}
```

Thus, your comment implies calling the `.get_abi()` method on the string type key in the dictionary, which would obviously not work.

fastlane_bot/events/pools/solidly_v2.py
```diff
@@ -85,15 +85,15 @@ async def update_from_contract(
...
         try:
-            factory_address = await contract.caller.factory()
+            factory_address = contract.caller.factory()
         except Exception:
             # Velocimeter does not expose factory function - call voter to get an address that is the same for all Velcoimeter pools
-            factory_address = await contract.caller.voter()
+            factory_address = contract.caller.voter()
 
-        self.is_stable = await contract.caller.stable()
+        self.is_stable = contract.caller.stable()
```
barakman, line 96: How are all of these changes related to the issue that this PR is attempting to address?

main.py
```diff
@@ -670,7 +671,13 @@ def run(
...
                     f"{len(mgr.pool_data)} total pools currently exist. Current block: {current_block}."
                 )
-                async_update_pools_from_contracts(mgr, current_block, logging_path)
+                try:
+                    async_update_pools_from_contracts(mgr, current_block, logging_path)
+                    failed_async_calls = 0
+                except Exception:
+                    failed_async_calls += 1
+                    if failed_async_calls >= 3:
+                        raise Exception(f"[main run.py] async_update_pools_from_contracts failed. List of failed pools: {mgr.pools_to_add_from_contracts}")
```
barakman, line 680: How are all of these changes related to the issue that this PR is attempting to address?

barakman (changes requested):

fastlane_bot/events/async_utils.py
barakman: Oh, heck, you're right.
The loop iterates the dictionary keys, not values.
So perhaps a better way to phrase it (though only for readability) would be:
```
for exchange_name, exchange in mgr.exchanges.items():
    abis[exchange_name] = exchange.get_abi()
```

main.py
Lesigh-3100: This fixes a different issue I noticed that caused trades including Solidly pools to fail, specifically when using the startup flag run_data_validator.

fastlane_bot/events/async_event_update_utils.py
```diff
@@ -536,3 +544,48 @@ def async_update_pools_from_contracts(mgr: Any, current_block: int, logging_path
...
         f"Async Updating pools from contracts took {(time.time() - start_time):0.4f} seconds"
     )
+
+
+def update_remaining_pools(mgr):
+    remaining_pools = []
+    all_events = [
+        event
+        for _, _, event, _, _ in mgr.pools_to_add_from_contracts
+    ]
```
barakman, line 554: More readable perhaps:
```
all_events = [pool[2] for pool in mgr.pools_to_add_from_contracts]
```
BTW, is each one of these "pools" guaranteed to always have at least 3 elements (or if you intend to keep your current version of the code - at least 5 elements)?

main.py
```diff
@@ -603,6 +602,7 @@ def run(
     last_block_queried = 0
     handle_static_pools_update(mgr)
     total_iteration_time = 0
+    failed_async_calls = 0
```
barakman, line 605: Where is this being used?

fastlane_bot/events/exceptions.py
```diff
@@ -4,4 +4,11 @@ def __init__(self, filepath):
 
     def __str__(self):
         return (f"tokens.csv does not exist at {self.filepath}. Please run the bot without the `read_only` flag to "
-                f"create this file.")
\ No newline at end of file
+                f"create this file.")
+
+
+class AyncUpdateRetryException(Exception):
```
barakman, line 10: Looks like there's a missing "s" at the beginning of this class name.

barakman (changes requested):
A lot of the changes in this PR seem cosmetic (spaces, newlines, etc), and some of them actually seem wrong, for example, in function `pair_name`).

If they're indeed wrong, then better not include them, so they don't distract us from the "real" changes.

fastlane_bot/events/exceptions.py
mikewcasale: This is a custom exception class

fastlane_bot/events/async_event_update_utils.py
mikewcasale: changed as you suggest

main.py
mikewcasale: removed as you suggest

fastlane_bot/events/async_utils.py
mikewcasale: I just implemented your suggestion.

mikewcasale (approved):

fastlane_bot/events/async_event_update_utils.py
```diff
@@ -12,11 +14,12 @@
 
 from fastlane_bot.data.abi import ERC20_ABI
 from fastlane_bot.events.async_utils import get_contract_chunks
-from fastlane_bot.events.exchanges import exchange_factory
+from fastlane_bot.events.exceptions import AyncUpdateRetryException
```
barakman, line 17: `AyncUpdateRetryException` seems to be missing an "s"

fastlane_bot/events/exceptions.py
barakman, line 10: `AyncUpdateRetryException` seems to be missing an "s"

barakman (approved):

fastlane_bot/events/async_event_update_utils.py
mikewcasale: this is a custom implementation

fastlane_bot/events/exceptions.py
mikewcasale: custom implementation - it's a child of Exception, no need for an s imo

PR #338, opened 2024-01-30
Add L1 gas fee calculation on Optimism & forks

Lesigh-3100:
This fix adds the calculation for "L1 Gas Fee" for transactions on Optimism & Optimism forks such as Base.

This gas fee is calculated & added to the current fee calculations to generate a much more accurate value for the expected transaction cost. 

Note:
Signing transactions has been moved outside the submit_transaction and submit_transaction_private functions to prevent signing the transaction twice as part of this flow.


fastlane_bot/helpers/txhelpers.py
```diff
@@ -375,13 +376,20 @@ def validate_and_submit_transaction(
...
+
         # Multiply expected gas by 0.8 to account for actual gas usage vs expected.
         gas_cost_eth = (
             Decimal(str(current_gas_price))
             * Decimal(str(gas_estimate))
             * Decimal(self.ConfigObj.EXPECTED_GAS_MODIFIER)
             / Decimal("10") ** Decimal("18")
         )
+        if self.ConfigObj.network.HAS_LAYER_ONE_GAS_FEE:
+            layer_one_gas_fee = asyncio.get_event_loop().run_until_complete(self.get_layer_one_gas_fee(signed_transaction=signed_arb_tx))
```
barakman, line 390: Looks pretty suspicious (`run_until_complete` in particularly).
Have you tested that?
Have you checked the runtime impact and the implications of this?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +972,57 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+            The total fee (in gas token) for the l1 gas fee
+        """
+        ethereum_base_fee = await self._get_layer_one_gas_price()
+        fixed_overhead = await self._get_layer_one_fee_overhead()
+        dynamic_overhead = await self._get_layer_one_fee_scalar()
+        zero_bytes, non_zero_bytes = count_bytes(signed_transaction["rawTransaction"])
+        tx_data_gas = zero_bytes * 4 + non_zero_bytes * 16
+        tx_total_gas = (tx_data_gas + fixed_overhead) * dynamic_overhead
+        l1_data_fee = tx_total_gas * ethereum_base_fee
+        l1_data_fee = Decimal(str(l1_data_fee)) / Decimal("1000000") / Decimal(str(10 ** 18))
```
barakman, line 999: I'm pretty sure that 'fee' is always denoted in wei units, which means that you should not be dividing by `10 ** 18`.
I also have no idea why you chose to divide by 1000000, which sounds related to the fee resolution on some DEXs (Carbon and Bancor, for example), but has absolutely nothing to do with gas fee in any way.
In short, it seems possible that both of these two divisions are not in place.
But if they were, then you may as well just divide by their multiplication, i.e., by `10 ** 24`.

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +972,57 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+        """
+        Returns the expected layer one gas fee for a layer 2 Optimism transaction
+        :param signed_transaction: the signed ethereum TX
+
+        returns: Decimal
+            The total fee (in gas token) for the l1 gas fee
+        """
+        ethereum_base_fee = await self._get_layer_one_gas_price()
+        fixed_overhead = await self._get_layer_one_fee_overhead()
+        dynamic_overhead = await self._get_layer_one_fee_scalar()
```
barakman, line 994: Consider running these async calls outside of this context (e.g., at the beginning of the loop), then you do not need to introduce any asynchronous-ity as part of this context (which seems to be forcing you to execute `run_until_complete` as part of this context).

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +972,57 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
                     self.ConfigObj.logger.info(
                         f"(***2***) Error when building transaction: {e.__class__.__name__} {e}")
             else:
                 return None
-        self.ConfigObj.logger.info(f"Submitting approval for token: {token_address}")
-        return self.submit_transaction(approve_tx)
 
+    async def get_layer_one_gas_fee(self, signed_transaction) -> Decimal:
+        """
+        Returns the expected layer one gas fee for a layer 2 Optimism transaction
```
barakman, line 986: Is this really unique to the `Optimism` chain?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +972,57 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+        tx_total_gas = (tx_data_gas + fixed_overhead) * dynamic_overhead
+        l1_data_fee = tx_total_gas * ethereum_base_fee
+        l1_data_fee = Decimal(str(l1_data_fee)) / Decimal("1000000") / Decimal(str(10 ** 18))
+        return l1_data_fee
+
+    async def _get_layer_one_gas_price(self):
+        """
+        Returns the layer one base fee from the Layer 2 gas oracle for Optimism blockchains
+        """
+        return self.ConfigObj.GAS_ORACLE_CONTRACT.caller.basefee()
```
barakman, line 1006: No need to append it with `.call()`?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +972,57 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+    async def _get_layer_one_fee_overhead(self):
+        """
+        Returns the layer one fee overhead from the Layer 2 gas oracle for Optimism blockchains
+        """
+        return self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeOverhead()
```
barakman, line 1011: No need to append it with `.call()`?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +972,57 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+    async def _get_layer_one_fee_scalar(self):
+        """
+        Returns the layer one fee overhead from the Layer 2 gas oracle for Optimism blockchains
+        """
+        return self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeScalar()
```
barakman, line 1016: No need to append it with `.call()`?

barakman (changes requested):

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: As far as I know - yes. (Including other Optimism deployments such as Base)

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: No this is correct

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: No this is correct

fastlane_bot/config/provider.py
```diff
@@ -18,7 +18,7 @@
 from ..data.abi import (
     BANCOR_V3_NETWORK_INFO_ABI,
     CARBON_CONTROLLER_ABI,
-    FAST_LANE_CONTRACT_ABI,
+    FAST_LANE_CONTRACT_ABI, GAS_ORACLE_ABI,
```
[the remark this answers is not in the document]
Lesigh-3100: Done

fastlane_bot/config/network.py
```diff
@@ -194,6 +194,7 @@ class ConfigNetwork(ConfigBase):
     # USDT_KEY = "USDT-1ec7"
     SELF_FUND = False
 
+    HAS_LAYER_ONE_GAS_FEE: bool
```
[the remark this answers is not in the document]
Lesigh-3100: Done

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +972,57 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
 
 
+def count_bytes(data: HexBytes) -> (int, int):
+    """
+    :param data: HexBytes
+        This is the HexBytes object from a signed TX
```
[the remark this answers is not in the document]
Lesigh-3100: Changed this

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +972,57 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+    returns Tuple(int, int):
+        The zero & non zero count of bytes in a transaction
```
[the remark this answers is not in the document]
Lesigh-3100: Changed this

resources/NBTest/NBTest_069_TestTxHelpers.py
```diff
@@ -0,0 +1,108 @@
...
+
+# +
+################### Base #####################
+cfg = Config.new(config=Config.CONFIG_MAINNET, blockchain="coinbase_base")
+
+tx_helpers = TxHelpers(cfg)
+
+def test_count_zero_bytes():
+    # Arrange
+    data = HexBytes(b'\x00\x00\x00\x00\x00')
```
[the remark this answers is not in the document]
Lesigh-3100: Updated to real data

fastlane_bot/helpers/txhelpers.py
barakman: I meant to say that in the context of this module, this information is irrelevant, even if it **currently** happens to be the case.

fastlane_bot/helpers/txhelpers.py
barakman: Well, the suggestion here is not just to add them, but rather - to add them only when needed.

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: I have tested that this works.

It's running inside the Main loop if that's what you mean? 

The issue is that if we need to make 3 contract calls to be able to calculate the L1 gas fee - if we don't use async it will be very slow. The function this is in is not asynchronous - so we need to use asyncio to call an async function.

If we put this in a different place, we will make additional network calls when they aren't necessary - and it won't speed up the cycle. 

If you have an alternative that you believe is faster I'm more than happy to discuss.

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Dividing by 10 ** 18:
The gas cost we calculated before this is in decimal format in order to compare it to the estimated profit in gas tokens. To add the l1 gas fee to it, it needs to be in the same format. 

Dividing by 1,000,000:
The l1FeeScalar returns a value in PPM, for example: 684000. So we need to multiply by 0.684. I chose to add the division at the end because everything until that last line is in int format. 

I'm changing this to divide by 10 ** 24

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Where could we add this that would be more efficient? 

Currently these are only being called when needed.

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: I think it's important to be specific here because otherwise it sounds like it could be relevant for layer one chains. Currently I'm only aware of Optimism using this model. If in the future more layer 2 solutions adopt this, we can clarify this further.

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: The function is taking around 1-1.18 seconds. If you have any ideas on how to speed it up I'm happy to discuss!

barakman: > If we put this in a different place, we will make additional network calls when they aren't necessary - and it won't speed up the cycle.

Not if you condition it in the same way it is conditioned here

---

> The function is taking around 1-1.18 seconds. If you have any ideas on how to speed it up I'm happy to discuss!

And do those 1-1.8 seconds take place only once per bot iteration?
Or do they occur within some other lower-level iteration, which means that they will repeat several times on every bot iteration)?

fastlane_bot/helpers/txhelpers.py
barakman: > for example: 684000. So we need to multiply by 0.684

Sorry, but this example makes zero sense.

The value of `684000 * 0.684` is `467856`. How exactly is that related to PPM resolution?

fastlane_bot/helpers/txhelpers.py
barakman: This file is a general helper utility.
It doesn't make sense to add specific-dex code or specific-chain code in here.
Same goes for documentation.
If this function is specific to Optimism, then it should move to the Optimism module.

fastlane_bot/helpers/txhelpers.py
barakman: I shall look into it...

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +974,58 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+                    return self.submit_transaction(approve_tx)
                 except Exception as e:
                     self.ConfigObj.logger.info(
                         f"(***2***) Error when building transaction: {e.__class__.__name__} {e}")
             else:
                 return None
-        self.ConfigObj.logger.info(f"Submitting approval for token: {token_address}")
-        return self.submit_transaction(approve_tx)
 
+    async def get_layer_one_gas_fee(self, signed_transaction) -> Decimal:
```
barakman, line 986: The name of a function which is not part of the module's interface (i.e., not designated to be called by other modules) should be prepended with an underscore

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +974,58 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+        """
+        ethereum_base_fee = await self._get_layer_one_gas_price()
+        fixed_overhead = await self._get_layer_one_fee_overhead()
+        dynamic_overhead = await self._get_layer_one_fee_scalar()
+        zero_bytes, non_zero_bytes = count_bytes(signed_transaction["rawTransaction"])
+        tx_data_gas = zero_bytes * 4 + non_zero_bytes * 16
+        tx_total_gas = (tx_data_gas + fixed_overhead) * dynamic_overhead
+        l1_data_fee = tx_total_gas * ethereum_base_fee
+        ## Dividing by 10 ** 24 because dynamic_overhead is returned in PPM format, and to convert this from WEI format to decimal format (10 ** 18).
+        l1_data_fee = Decimal(str(l1_data_fee)) / Decimal(str(10 ** 24))
```
barakman, line 1002: Can use:
```
l1_data_fee = Decimal(f"{l1_data_fee}e-24")
```
Or even:
```
return Decimal(f"{l1_data_fee}e-24")
```

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +974,58 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+    async def _get_layer_one_fee_overhead(self):
+        """
+        Returns the layer one fee overhead from the Layer 2 gas oracle for Optimism blockchains
+        """
+        return self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeOverhead()
+    async def _get_layer_one_fee_scalar(self):
+        """
+        Returns the layer one fee overhead from the Layer 2 gas oracle for Optimism blockchains
+        """
+        return self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeScalar()
```
barakman, line 1020: These 3 functions are one line each; I'd move that directly to where you call them; will clear up the code

fastlane_bot/helpers/txhelpers.py
```diff
@@ -969,13 +974,58 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
+def count_bytes(data: HexBytes) -> (int, int):
+    """
+    This function counts the number of zero and non-zero bytes in a given input data.
+    :param data: HexBytes
+    returns Tuple(int, int):
+        The zero & non zero count of bytes in the input
+    """
+    zero_bytes = len([byte for byte in data if byte == 0])
+    non_zero_bytes = len(data) - zero_bytes
+    return zero_bytes, non_zero_bytes
```
barakman, line 1031: This looks like a general-utility function.
Consider moving it to the general utility module, if it makes sense for it to be there.

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Done

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Moved to utils.py

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Done

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Added underscore

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: It doesn't need to occur once per bot iteration - it's only necessary when we get to this point in the code. It will only occur once per cycle.

barakman (approved):

PR #339, opened 2024-01-30
337 failed to get tokens and fee for aerodrome v2

Lesigh-3100:
Initialized Solidly Factory contracts using web3 async instead of web3, which was causing Solidly pools to fail to fetch fees in the async function.

### DIFF
--- a/fastlane_bot/events/managers/base.py
+++ b/fastlane_bot/events/managers/base.py
@@ -134,12 +134,11 @@ class BaseManager:
         Handles getting stable & volatile fees for Solidly forks
         """
         exchange_name = exchange.exchange_name
-        self.factory_contracts[exchange_name] = self.web3.eth.contract(
+        self.factory_contracts[exchange_name] = self.w3_async.eth.contract(
             address=self.cfg.FACTORY_MAPPING[exchange_name],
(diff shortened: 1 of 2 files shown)


mikewcasale (approved):

fastlane_bot/events/managers/base.py
```diff
@@ -134,12 +134,11 @@ def handle_solidly_exchanges(self, exchange):
         Handles getting stable & volatile fees for Solidly forks
         """
         exchange_name = exchange.exchange_name
-        self.factory_contracts[exchange_name] = self.web3.eth.contract(
+        self.factory_contracts[exchange_name] = self.w3_async.eth.contract(
```
barakman, line 137: This change might lead to the following problem:

In the bancor_v3 exchange, [function get_fee is not declared 'async'](https://github.com/bancorprotocol/fastlane-bot/blob/main/fastlane_bot/events/exchanges/bancor_v3.py#L41), however, it is [called with 'await' for ALL exchanges](https://github.com/bancorprotocol/fastlane-bot/blob/main/fastlane_bot/events/async_event_update_utils.py#L72):
```
fee = await ex.get_fee(address, contract)
```
Which means that this function probably has to be declared `async`.

PR #342, opened 2024-01-30
bugfix for excessive main loop runtime

mikewcasale:


zavelevsky (approved):
LGTM

fastlane_bot/events/async_event_update_utils.py
```diff
@@ -544,45 +544,3 @@ def async_update_pools_from_contracts(mgr: Any, current_block: int, logging_path
...
-            mgr.cfg.logger.warning("[run_async_update_with_retries] ex_name not found from event")
-            continue
-
-        key, key_value = mgr.get_key_and_value(event, addr, ex_name)
-        pool_info = mgr.get_pool_info(key, key_value, ex_name)
-
-        if not pool_info:
-            remaining_pools.append((addr, ex_name, event, key, key_value))
-
-    shuffle(remaining_pools)  # shuffle to avoid repeated immediate failure of the same pool
```
barakman, line 565: The `import shuffle` statement is probably redundant in this file by now

fastlane_bot/events/managers/manager.py
```diff
@@ -360,3 +360,21 @@ def handle_trading_fee_updated(self):
...
+                self.cfg.logger.warning("[run_async_update_with_retries] ex_name not found from event")
+                continue
+
+            key, key_value = self.get_key_and_value(event, addr, ex_name)
+            pool_info = self.get_pool_info(key, key_value, ex_name)
+
+            if not pool_info:
+                remaining_pools.append((addr, ex_name, event, key, key_value))
+
+        return remaining_pools
```
barakman, line 380: No need to return `remaining_pools`, just use it for updating `self.pools_to_add_from_contracts` in this function

main.py
```diff
@@ -854,6 +857,29 @@ def run(
...
+
+    while failed_async_calls < max_retries:
+        try:
+            async_update_pools_from_contracts(mgr, current_block, logging_path)
+            return  # Successful execution
+        except AyncUpdateRetryException as e:
+            failed_async_calls += 1
+            remaining_pools = mgr.update_remaining_pools()
+            shuffle(remaining_pools)  # shuffle to avoid repeated immediate failure of the same pool
+            mgr.pools_to_add_from_contracts = remaining_pools
```
barakman, line 871: In conjunction with a previous comment,  function `mgr.update_remaining_pools` can handle this assignment internally

main.py
```diff
@@ -854,6 +857,29 @@ def run(
...
+    failed_async_calls = 0
+
+    while failed_async_calls < max_retries:
+        try:
+            async_update_pools_from_contracts(mgr, current_block, logging_path)
+            return  # Successful execution
+        except AyncUpdateRetryException as e:
+            failed_async_calls += 1
+            remaining_pools = mgr.update_remaining_pools()
+            shuffle(remaining_pools)  # shuffle to avoid repeated immediate failure of the same pool
```
barakman, line 870: Why is shuffling applied only starting from the first failed attempt?

Sounds like it should be applied on the original list of pools, rather than (or in addition to) being applied on the list of remaining pools upon every retry.

barakman (changes requested):

fastlane_bot/events/async_event_update_utils.py
mikewcasale: fixed

fastlane_bot/events/managers/manager.py
mikewcasale: fixed

main.py
mikewcasale: fixed

main.py
mikewcasale: The assumption is that pools will not fail - this is true ~99% of the time (ie that the first attempt here will succeed). Only if while iterating through the list, one of the pools fails, and then it retries to process any remaining non-problematic pools, does shuffling become sensible. Otherwise it is just an extra operation.

main.py
```diff
@@ -854,6 +856,27 @@ def run(
...
+    failed_async_calls = 0
+
+    while failed_async_calls < max_retries:
+        try:
+            async_update_pools_from_contracts(mgr, current_block, logging_path)
+            return  # Successful execution
+        except AyncUpdateRetryException as e:
+            failed_async_calls += 1
+            mgr.update_remaining_pools()
+            mgr.cfg.logger.error(f"Attempt {failed_async_calls} failed: {e}")
```
barakman, line 869: Perhaps move this logging message to be printed prior to calling the `update_remaining_pools` function, in case this function itself throws an exception.

barakman (changes requested):

barakman (approved):

main.py
mikewcasale: fixed

PR #345, opened 2024-02-01
Clean up Solidly fee calculations

Lesigh-3100:
Fixed Velocimeter Fee calculation + small code cleanup.

### DIFF
--- a/fastlane_bot/events/exchanges/solidly_v2.py
+++ b/fastlane_bot/events/exchanges/solidly_v2.py
@@ -38,11 +38,6 @@ class SolidlyV2(Exchange):
         return float(self.fee)
 
     def add_pool(self, pool: Pool):
-
-        # pool_fee = self._get_pool_fee(pool=pool)
-        # pool.fee = pool_fee
-        # pool.state["fee"] = pool_fee
-        # pool.state["fee_float"] = pool_fee
         self.pools[pool.state["address"]] = pool
 
     async def _get_pool_fee(self, pool: Pool):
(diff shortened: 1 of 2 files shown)


fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -61,20 +56,15 @@ def get_events(self, contract: Contract) -> List[Type[Contract]]:
     async def get_fee(self, address: str, contract: Contract) -> Tuple[str, float]:
         if "velocimeter" in self.exchange_name:
             default_fee = await self.factory_contract.caller.getFee(address)
-            default_fee = float(default_fee)
-            if default_fee <= 1000:
-                default_fee = default_fee / 10000
-            else:
-                default_fee = default_fee / 10 ** 18
+            default_fee = float(default_fee) / 10000
```
barakman, line 59: This doesn't reconcile with function `get_pool_fee`, which divides by `10 ** 18`

run_blockchain_terraformer.py
```diff
@@ -683,20 +683,13 @@ def organize_pool_details_solidly_v2(
...
     elif "scale" in exchange:
         default_fee = factory_contract.caller.getRealFee(pool_address)
-        default_fee = float(default_fee)
-        default_fee = default_fee / 10 ** 18
+        default_fee = float(default_fee) / 10 ** 18
     else:
         default_fee = factory_contract.caller.getFee(pool_address, is_stable)
-        default_fee = float(default_fee)
-        default_fee = default_fee / 10000
+        default_fee = float(default_fee) / 10000
```
barakman, line 692: Why do we have the same code replicated in two different places?

barakman (changes requested):

barakman:
As part of the small code cleanup, you may get rid of function `set_stable_volatile_fee` if it's not used anywhere.

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: Thank you - that was a leftover function from development that I forgot to remove. I've removed get_pool_fee and a few other unused functions.

run_blockchain_terraformer.py
Lesigh-3100: Between solidly_v2 and the terraformer? If that's the question - it's because the terraformer was intended to be run independently. 

If the question is why do we call getFee twice - it's because the functions have different inputs.

barakman: No, my point was that we should use the same function in both places rather than replicating it

Lesigh-3100: I see your point, but it would require a major refactor of the terraformer file.

barakman (approved):

Issue #348, opened 2024-02-04
Minor optimization in multicaller


Lesigh-3100:
## Enhancement Description
Minor optimization to multicaller class - avoid reinitialization of multicall contract. 

## Rationale
This upgrade reuses the multicall contract instead of recreating it multiple times. This is a small change that results in a performance boost.

## Impact Analysis
- Automated Tests: Implications on testing (new/changed)
None
- Risks: What are the risks associated with this enhancement?
Can't think of anything.

- Performance: Expected performance impacts and how to measure them
This improves performance when using multicall. This is can be seen in the avg time per iteration. 

- Monitoring: Changes required in the monitoring system
None
- Logging: Do we need additional logs?
No
- Documentation: Updates or new documentation needed?
No

barakman:
This issue cannot be explained as the result of calling `web3.eth.contract(...)` twice instead of once at some place in the code, because the time it takes to create the logical object which represents the multi-call contract is approximately 0.01 of a second (see python script below).

If there is indeed a notable performance improvement as a result of fixing the above, then it implies either one of two options:
1. The place in the code where this logical object is created, is called repeatedly from somewhere else in the code
2. The performance measurement itself is biased due to a "fluctuating" (inconsistent) environment-state

In order to investigate option 1 properly, one needs to follow every execution path which leads to that point in the code, and verify that it is not unnecessarily repeated over and over.

This can be done by searching for occurrences of `multicall(` in the code (case-sensitive AND whole-word), since that's the function where the logical object is created twice instead of once.

Here are some of the different execution paths in which that function is called:

1:
```
multicall_helper -->
    multicall_every_iteration -->
       run (called once per iteration)
```
2:
```
get_strats_by_contract -->
    get_strategies -->
        update_carbon -->
            get_rows_to_update -->
                async_handle_initial_iteration -->
                   run (called once per iteration)
```
3:
```
get_strats_by_contract -->
    get_strategies -->
        update_carbon -->
            get_rows_to_update -->
                handle_initial_iteration (not called anywhere)
```
4:
```
get_fees_by_pair -->
    get_fee_pairs -->
        set_carbon_v1_fee_pairs -->
            BaseManager.__post_init__ -->
                main (called once per execution)
```
5:
```
get_fees_by_pair -->
    get_fee_pairs -->
        update_carbon -->
            get_rows_to_update -->
                async_handle_initial_iteration -->
                   run (called once per iteration)
```
6:
```
[...]

PR #349, opened 2024-02-04
Avoid contract reinitialization in multicaller

Lesigh-3100:
Issue:
https://github.com/bancorprotocol/fastlane-bot/issues/348

This change initializes the Multicall contract a single time and resuses it instead of reinitializing.

### DIFF
--- a/fastlane_bot/config/multicaller.py
+++ b/fastlane_bot/config/multicaller.py
@@ -117,6 +117,10 @@ class MultiCaller(ContextManager):
         self.block_identifier = block_identifier
         self.web3 = web3
         self.MULTICALL_CONTRACT_ADDRESS = self.web3.to_checksum_address(multicall_address)
+        self.MULTICALL_CONTRACT = self.web3.eth.contract(
+                abi=MULTICALL_ABI,
+                address=self.MULTICALL_CONTRACT_ADDRESS
+            )
 
     def __enter__(self) -> 'MultiCaller':
         return self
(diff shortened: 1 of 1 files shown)


barakman (commented):
This fix doesn't really do anything other than (semantically) creating the `self.MULTICALL_CONTRACT` logical object once instead of twice.

barakman (commented):
Posted by mistake

barakman (changes requested):
This fix doesn't really do anything other than semantically creating the `self.MULTICALL_CONTRACT` logical object once instead of twice.
There is most likely very little added value to it, in the form of a negligible performance improvement.

It does, however, put in doubt the first usage of this object, which seems to be merely creating `_encoded_data` list, and then throwing an exception if the first element in that list does not adhere a certain condition.

Following that assertion, the entire `_encoded_data` list is discarded (no longer in use).
If that's the case, then why generate this entire list and not just the first element in it?

Following that, a new object named `encoded_data` is generated in an almost identical manner.
Is it possible that you may as well use the previous object instead of this one?
Or perhaps generate only the new object, assert its first element and then proceed with it?

Lesigh-3100:
> This fix doesn't really do anything other than semantically creating the `self.MULTICALL_CONTRACT` logical object once instead of twice. There is most likely very little added value to it, in the form of a negligible performance improvement.
> 
> It does, however, put in doubt the first usage of this object, which seems to be merely creating `_encoded_data` list, and then throwing an exception if the first element in that list does not adhere a certain condition.
> 
> Following that assertion, the entire `_encoded_data` list is discarded (no longer in use). If that's the case, then why generate this entire list and not just the first element in it?
> 
> Following that, a new object named `encoded_data` is generated in an almost identical manner. Is it possible that you may as well use the previous object instead of this one? Or perhaps generate only the new object, assert its first element and then proceed with it?

Initializing contracts with Web3 is actually fairly slow, so this makes a difference. Ideally it should only be initialized once, possibly stored in Manager, and reused from there. 

I see your point regarding  _encoded_data - I'm going to look into removing it.

barakman:
> Initializing contracts with Web3 is actually fairly slow, so this makes a difference. Ideally the Multicaller class should only be initialized once, possibly stored in Manager, and reused from there.

AFAIK, this is not true.

The object created via `web3.eth.contract(...)` is a **logical** object, "purely offchain" if you will, which represents the contract deployed onchain. Its purpose is to allow for a cleaner interaction between your offchain code and the onchain contract, i.e., a "class-like" interaction, where can call the contract functions without having to worry about the underlying web3 communication with the remote node, etc.

Creating this object should take literally 2 milliseconds, so reducing the code from creating it twice to creating it once, isn't supposed to yield any significant improvement in the bot's performance. If it does, then that would imply that the creation of that object is being "repeated" somehow from an outer loop or something.

You should make sure to run the comparison within an isolated manner, where EVERYTHING except for this one change remains identical.

For example, if you're running the bot on a real chain, then with the state of the chain constantly changing, performance comparison becomes completely invalid.

In fact, even some data which is cached on your local machine between your measurement of the old code and your measurement of the new code, might impact the results.

Issue #353, opened 2024-02-05
bug in _get_layer_one_gas_fee


Lesigh-3100:
## Bug Description
In the function _get_layer_one_gas_fee, contract call fail as currently implemented & must be wrapped in async functions.

## Severity (High/Medium/Low)
High

## Steps to Reproduce
Run bot on Base. Bot will fail when it tries to get L1 gas fee. 
## Expected Behavior
Bot should get L1 gas fee.

## Actual Behavior
Bot fails to get L1 gas fee.

## Impact Analysis (to be filled by contributors)
- Automated Tests: Are there existing tests covering this? Do we need new tests?
The API calls could be mocked - but once this is working it is unlikely to change until the calculation is updated. 

- Risks: Potential areas affected by this bug and its probable fix
This is a critical bug. It should be verified that it is functioning correctly before merging. 
- Performance: Is this bug affecting performance?
No
- Monitoring: Does this affect any current monitoring? Any new alerts needed?
No
- Logging: Any changes required in logging?
No
- Documentation: Updates needed in documentation?
No

PR #354, opened 2024-02-05
Update txhelpers.py

Lesigh-3100:
Issue:
https://github.com/bancorprotocol/fastlane-bot/issues/353

The bot currently throws the following error in the function _get_layer_one_gas_fee:
[bot:run:single] object int can't be used in 'await' expression

Wrapping the contract calls in async functions & awaiting those functions fixes this.

### DIFF
--- a/fastlane_bot/helpers/txhelpers.py
+++ b/fastlane_bot/helpers/txhelpers.py
@@ -992,8 +992,11 @@ class TxHelpers:
             The total fee (in gas token) for the l1 gas fee
         """
 
-        tasks = self.ConfigObj.GAS_ORACLE_CONTRACT.caller.basefee(), self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeOverhead(), self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeScalar()
-        ethereum_base_fee, fixed_overhead, dynamic_overhead = asyncio.get_event_loop().run_until_complete(asyncio.gather(*tasks))
+        ethereum_base_fee, fixed_overhead, dynamic_overhead = asyncio.get_event_loop().run_until_complete(asyncio.gather(
+            self.ConfigObj.GAS_ORACLE_CONTRACT.caller.basefee(),
+            self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeOverhead(),
+            self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeScalar()
+        ))
 
         return self._get_layer_one_gas_fee(signed_transaction=signed_transaction, ethereum_base_fee=ethereum_base_fee, fixed_overhead=fixed_overhead, dynamic_overhead=dynamic_overhead)
 
--- a/fastlane_bot/helpers/txhelpers.py
+++ b/fastlane_bot/helpers/txhelpers.py
@@ -991,41 +991,28 @@ class TxHelpers:
         returns: Decimal
             The total fee (in gas token) for the l1 gas fee
         """
-        return asyncio.get_event_loop().run_until_complete(self._get_layer_one_gas_fee(signed_transaction=signed_transaction))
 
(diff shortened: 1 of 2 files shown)


fastlane_bot/helpers/txhelpers.py
```diff
@@ -1001,15 +1001,31 @@ async def _get_layer_one_gas_fee(self, signed_transaction) -> Decimal:
         returns: Decimal
             The total fee (in gas token) for the l1 gas fee
         """
-        ethereum_base_fee = await self.ConfigObj.GAS_ORACLE_CONTRACT.caller.basefee()
-        fixed_overhead = await self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeOverhead()
-        dynamic_overhead = await self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeScalar()
+        ethereum_base_fee = await self._get_layer_one_gas_price()
+        fixed_overhead = await self._get_layer_one_fee_overhead()
+        dynamic_overhead = await self._get_layer_one_fee_scalar()
```
barakman, line 1006: I believe that returning all values from a single (async) function would be more efficient (otherwise, you're essentially running these functions in a sequential manner, awaiting for the completion of each function-call before moving to the next).

barakman (changes requested):

fastlane_bot/helpers/txhelpers.py
```diff
@@ -991,19 +991,22 @@ def _get_layer_one_gas_fee_loop(self, signed_transaction) -> Decimal:
         returns: Decimal
             The total fee (in gas token) for the l1 gas fee
         """
-        return asyncio.get_event_loop().run_until_complete(self._get_layer_one_gas_fee(signed_transaction=signed_transaction))
 
-    async def _get_layer_one_gas_fee(self, signed_transaction) -> Decimal:
+        tasks = self.ConfigObj.GAS_ORACLE_CONTRACT.caller.basefee(), self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeOverhead(), self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeScalar()
+        ethereum_base_fee, fixed_overhead, dynamic_overhead = asyncio.get_event_loop().run_until_complete(asyncio.gather(*tasks))
```
barakman, line 996: Cleaner IMO (instead of creating an array of tasks and then dereferencing it with `*`):
```
ethereum_base_fee, fixed_overhead, dynamic_overhead = asyncio.get_event_loop().run_until_complete(
    asyncio.gather([
        self.ConfigObj.GAS_ORACLE_CONTRACT.caller.basefee(),
        self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeOverhead(),
        self.ConfigObj.GAS_ORACLE_CONTRACT.caller.l1FeeScalar()
    ])
)
```
Also, please search if there's a function which achieves what `run_until_complete(asyncio.gather(...))`.
Make sense that there should be a single function which takes care of that.

fastlane_bot/helpers/txhelpers.py
```diff
@@ -991,19 +991,22 @@ def _get_layer_one_gas_fee_loop(self, signed_transaction) -> Decimal:
...
+
+        return self._get_layer_one_gas_fee(signed_transaction=signed_transaction, ethereum_base_fee=ethereum_base_fee, fixed_overhead=fixed_overhead, dynamic_overhead=dynamic_overhead)
```
barakman, line 998: Semantic:
No need to specify each parameter name as part of this function-call.
This syntax is only really needed for input arguments which have default values in the called function.
Using it here makes this function-call statement unnecessarily long, and rather difficult to read.

barakman (changes requested):

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Updated this.

barakman (approved):

mikewcasale (approved):

Issue #352, opened 2024-02-05
Revise all submit-transaction methods in the TxHelpers module


barakman:
The following methods are used for submitting transactions:
1. `submit_transaction`
2. `submit_private_transaction`
3. `cancel_private_transaction`

These methods are used inconsistently across the code, sometimes receiving a signed transaction as input, sometimes receiving an unsigned transaction as input, and sometimes receiving a raw transaction as input.

In addition to that, not every attempt to submit a transaction is complemented with an attempt to cancel it upon timeout.
This requires a higher-level decision on which transactions should be cancelled when needed, and which should not be.

Technically, submitting a transaction requires only the signed transaction as input, but cancelling it requires the original (unsigned) transaction which was used in order to create it, as input.

So designing a solution for this problem should also take into account how every submitted transaction which may need to be cancelled, can be easily cancelled.

An addition bug in the cancellation method, is that it resets the transaction's `data` attribute but not its `value` attribute.
As a result, cancelling a transaction which includes ETH, might end up with that transaction being only "partially cancelled" (namely, the ETH-transfer part in that transaction will effectively take place despite the cancellation).

The aforementioned functions are used only in one place outside of this module, in method `generate_shutdown_tx`.
So the implemented solution should also include this specific usage.

barakman:
Handled in PR #355

PR #355, opened 2024-02-05
Revise all submit-transaction methods in the TxHelpers module

barakman:
Handles issue #352

### DIFF
--- a/fastlane_bot/helpers/txhelpers.py
+++ b/fastlane_bot/helpers/txhelpers.py
@@ -378,7 +378,6 @@ class TxHelpers:
 
         signed_arb_tx = self.sign_transaction(arb_tx)
 
-
         # Multiply expected gas by 0.8 to account for actual gas usage vs expected.
         gas_cost_eth = (
             Decimal(str(current_gas_price))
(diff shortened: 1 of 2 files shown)


fastlane_bot/helpers/txhelpers.py
```diff
@@ -744,109 +737,70 @@ def build_tx(
...
+
+    def _submit_transaction(self, tx_hash) -> str:
+        try:
+            tx_receipt = self.web3.eth.wait_for_transaction_receipt(tx_hash)
+            assert tx_hash == tx_receipt["transactionHash"]
+            return tx_hash
+        except TimeExhausted as e:
             self.ConfigObj.logger.info(
-                f"[helpers.txhelpers.cancel_private_transaction] Failed to submit cancellation to Flashbots RPC"
+                f"[helpers.txhelpers._submit_transaction] Transaction stuck in mempool for 120 seconds, cancelling."
```
Lesigh-3100, line 803: Can we change this message to:
f"[helpers.txhelpers._submit_transaction] Transaction stuck in mempool for 120 seconds, moving on."

Since we are no longer submitting a cancellation.

barakman: Ah yeah, that's right.
We should probably also just use "timeout" instead of "120 seconds" because this is largely depends on the internal implementation of web3.py and/or the remote node in use.

barakman: Fixed

zavelevsky (approved):
LGTM

mikewcasale (approved):

PR #359, opened 2024-02-06
Update static data for Base, fix Sushi V3 Factory address

mikewcasale:


barakman (approved):

run_blockchain_terraformer.py
```diff
@@ -702,7 +703,7 @@ def organize_pool_details_solidly_v2(
         return None
     description = exchange + " " + pair
 
-
+    exchange_id = 12 if exchange == AERODROME_V2_NAME else 11
```
barakman, line 706: Please find a way to reuse the values `11` and `12` rather than having each one of them hard-coded in (at least) two different places.
Also, this assignment is generally unclear, as in implies that the value of `exchange` must be either "velocimeter" or "solidly", in which case, it should be made clear that these are the only two expected values.
If it is not the case, then note that this line forces it to be (which will probably result in undesired behavior.

barakman (changes requested):

barakman (approved):

Issue #346, opened 2024-02-01
Add Handling of Flashloan Fee


Lesigh-3100:
## Enhancement Description
This feature will add support for flashloans from sources that take a fee. 

The implementation will be to add a flashloan fee constant per blockchain in network.py, and use it to calculate the expected profit after flashloan fee.

## Rationale
- Why is this enhancement important?
This is important to support blockchains in which there is no viable free flashloan source. 

## Impact Analysis
- Automated Tests: Implications on testing (new/changed)
Add Test to check that the fee is working for each chain. Add unit test for the calculation. 

- Risks: What are the risks associated with this enhancement?
No obvious risks. 

- Performance: Expected performance impacts and how to measure them
No expected performance impact.
- Monitoring: Changes required in the monitoring system
Nothing
- Logging: Do we need additional logs?
Optional log stating expected flashloan fee
- Documentation: Updates or new documentation needed?
No

ashachaf:
consider a case in which you have more than 1 source of flashloan and each source and token might have a different fee level

PR #360, opened 2024-02-07
Adds support for custom flashloan fees

mikewcasale:
Handles [issue 346](https://github.com/bancorprotocol/fastlane-bot/issues/346)


fastlane_bot/config/constants.py
```diff
@@ -0,0 +1,13 @@
...
+
+(c) Copyright Bprotocol foundation 2023.
+Licensed under MIT License.
+"""
+
+FLASHLOAN_FEE_MAP = {
+    # should be in the form of {network: fee * 1e6}
+    "ethereum": 0,
+    "coinbase_base": 0,
+    "fantom": 300,  # Flashloans come with a 0.03% Fee from Beethoven X (Balancer fork)
```
zavelevsky, line 12: Why represent percentage like that?

mikewcasale: Is this not a standard approach in the industry, to specify in PPM? In any case, just following what we did in the simulator, and as well how Uniswap does it, etc. Happy to change if there is a more preferred method.

barakman: 1. It is not standard, but an implementation-decision (typically per DEX).
2. Unless this is related to something the you need to process before passing as input to a contract function or after receiving as output from a contract function, there is no reason for you to use this resolution to begin with; the only reason for it being used onchain, is the lack of non-integer-arithmetic support, which is obviously not an issue in offchain scripts.

fastlane_bot/bot.py
```diff
@@ -803,15 +808,16 @@ def calculate_profit(
             self.ConfigObj.logger.debug(f"[bot.calculate_profit sorted_price_curves] {sorted_price_curves}")
             if len(sorted_price_curves)>0:
                 fltkn_eth_conversion_rate = sorted_price_curves[0][-1]
-                best_profit_eth = Decimal(str(best_profit_fl_token)) / Decimal(str(fltkn_eth_conversion_rate))
+                flashloan_fee_amt_eth = Decimal(str(flashloan_fee_amt_fl_token)) / Decimal(str(fltkn_eth_conversion_rate))
+                best_profit_eth = Decimal(str(best_profit_fl_token)) / Decimal(str(fltkn_eth_conversion_rate)) - flashloan_fee_amt_eth
```
barakman, line 812: Possibly better to avoid dividing by the ETH conversion rate here, and leave it to take place (if really necessary) at a different point during execution?

If yes, then the two lines above become somewhat simpler, cleaner, and most importantly - more accurate:
```
flashloan_fee_amt_eth = flashloan_fee_amt_fl_token
best_profit_eth = best_profit_fl_token - flashloan_fee_amt_fl_token
```

fastlane_bot/bot.py
```diff
@@ -803,15 +808,16 @@ def calculate_profit(
...
                 self.ConfigObj.logger.debug(f"[bot.calculate_profit] {fl_token, best_profit_fl_token, fltkn_eth_conversion_rate, best_profit_eth, 'ETH'}")
             else:
                 self.ConfigObj.logger.error(
                     f"[bot.calculate_profit] Failed to get conversion rate for {fl_token} and {self.ConfigObj.WRAPPED_GAS_TOKEN_ADDRESS}. Raise"
                 )
                 raise
         else:
-            best_profit_eth = best_profit_fl_token
+            best_profit_eth = best_profit_fl_token - flashloan_fee_amt_fl_token
```
barakman, line 820: Oh, I see that my suggestion above is actually applied here, so the question arises - why are you **currently** calculating `best_profit_eth` in two different ways (dividing by the ETH conversion rate in one place, and not diving by the ETH conversion rate in another place)?

fastlane_bot/helpers/routehandler.py
```diff
@@ -501,13 +502,13 @@ def get_custom_address(
             return pool.tkn0_address
 
 
-    def generate_flashloan_struct(self, trade_instructions_objects: List[TradeInstruction]) -> List:
+    def generate_flashloan_struct(self, trade_instructions_objects: List[TradeInstruction]) -> Tuple[List, int]:
```
barakman, line 505: This function does nothing but calling `self._get_flashloan_struct`, so you may as well just get rid of it and rename the other one to `generate_flashloan_struct`.

barakman (changes requested):

fastlane_bot/helpers/routehandler.py
mikewcasale: Yes, I am aware of it, but it is unrelated to the changes in the PR.

fastlane_bot/bot.py
mikewcasale: This is why it helps to look at more than just the diffs. If the flashloan is not in ETH/WETH, then it is converted into ETH with:
```python
flashloan_fee_amt_eth = Decimal(str(flashloan_fee_amt_fl_token)) / Decimal(str(fltkn_eth_conversion_rate))
```

otherwise, if the flashloan is in ETH/WETH, then:
```python
best_profit_eth = best_profit_fl_token - flashloan_fee_amt_fl_token
```

Because of this, I'm not really sure where _the other place_ you suggest above would be? (ie the value is used here.

fastlane_bot/config/constants.py
mikewcasale: In this case, what is your preferred implementation approach / alternative?

barakman: Why not just use `0.0003` (i.e., multiply by that value wherever needed)?

zavelevsky: I'd go with the most obvious - either `0.0003 # 0.03%` or `0.03 # %`

fastlane_bot/helpers/routehandler.py
```diff
@@ -529,7 +530,7 @@ def _get_flashloan_platform_id(self, tkn: str) -> int:
         else:
             return 7
 
-    def _get_flashloan_struct(self, trade_instructions_objects: List[TradeInstruction]) -> List:
+    def _get_flashloan_struct(self, trade_instructions_objects: List[TradeInstruction]) -> Tuple[List, int]:
```
zavelevsky, line 533: where do you actually return the fee?

mikewcasale: Should be corrected now.

fastlane_bot/config/constants.py
mikewcasale: This has been updated.

fastlane_bot/bot.py
mikewcasale: I am pretty sure I already responded to this comment somewhere but dont see it now... In any case, because the profit is calculated and returned within this function, I dont know why we would (or should) treat the flashloan fee any different than the rest of the profit calculation(s)?

Once the calculated profit leaves this function it is compared against the gas costs and the user-specified minimum profit (in simple great-than, less-than comparisons). IMO the most intuitive thing to do is keep the profit calculation contained within a single function like it is now, and not delay part of the calculation to happen elsewhere.

fastlane_bot/bot.py
```diff
@@ -964,14 +969,17 @@ def _handle_trade_instructions(
         # Get the flashloan token
         fl_token = calculated_trade_instructions[0].tknin_address
         fl_token_symbol = calculated_trade_instructions[0].tknin_symbol
+        flashloan_amount = int(calculated_trade_instructions[0].amtin_wei)
+        flashloan_fee = FLASHLOAN_FEE_MAP.get(self.ConfigObj.NETWORK, 0)
+        flashloan_fee_amt = int(flashloan_fee * flashloan_amount)
```
barakman, line 974: In the initialization of `FLASHLOAN_FEE_MAP`, you are using a percentage value of 0.03% (at least by the comment next to it), i.e., a value between 0 and 100.
But the calculation in the line above implies that you are using a normalized value (between 0 and 1).
So either change the map to consist of normalized values, or add in this line a division by 100.

fastlane_bot/config/constants.py
```diff
@@ -0,0 +1,13 @@
...
+
+(c) Copyright Bprotocol foundation 2023.
+Licensed under MIT License.
+"""
+
+FLASHLOAN_FEE_MAP = {
+    # should be in the form of {network: fee * 1e6}
+    "ethereum": 0,
+    "coinbase_base": 0,
+    "fantom": 0.03  # %
```
barakman, line 12: My previous comment assumes that 0.03 here stands for 0.03%.

barakman (changes requested):

fastlane_bot/config/constants.py
barakman: An important note to make is that Mike's original implementation to use a PPM factor, was actually a good one in the sense of avoiding floating-point calculations as much as possible (i.e., the calculation would subsequently turn into `amount * fee // 1e6`, which is purely integer-arithmetic).
But not critical at this point, especially after he's already changed that.

Issue #361, opened 2024-02-07
Add support for custom RPC


mikewcasale:
## Enhancement Description
- Add support for setting a custom (i.e. non-alchemy provider RPC).

## Rationale
- Can be useful in a number of situations, for example: (1) when running the bot on a blockchain that is not supported by Alchemy, when running a custom node, &etc.

## Impact Analysis
- Automated Tests: Implications on testing - None known.
- Risks: What are the risks associated with this enhancement? - If the bot connects to an unintended provider this could produce unexpected behavior.
- Performance: Expected performance impacts and how to measure them - None known.
- Monitoring: Changes required in the monitoring system - None
- Logging: Do we need additional logs? - Yes, additional logging records should indicate the provider.
- Documentation: Updates or new documentation needed? - In main.py a new `rpc_url` option should have a brief description, and usage example.

PR #362, opened 2024-02-07
Adds custom RPC url support

mikewcasale:
Handles issue #361


fastlane_bot/config/connect.py
```diff
@@ -18,6 +18,14 @@
...
 
 from fastlane_bot.config import network
+from fastlane_bot.config.constants import ARBITRUM_ONE_NETWORK, COINBASE_BASE_NETWORK, ETHEREUM_NETWORK, \
+    OPTIMISM_NETWORK, POLYGON_NETWORK, \
+    POLYGON_ZKEVM_NETWORK, \
+    WEB3_ALCHEMY_ARBITRUM, WEB3_ALCHEMY_BASE, \
+    WEB3_ALCHEMY_OPTIMISM, \
+    WEB3_ALCHEMY_POLYGON, \
+    WEB3_ALCHEMY_POLYGON_ZKEVM, \
+    WEB3_ALCHEMY_PROJECT_ID
```
barakman, line 28: Semantic: I think that you can get rid of the backslashes by placing all imported symbols inside `()`.

fastlane_bot/config/connect.py
```diff
@@ -79,6 +87,35 @@ def __init__(
...
+            rpc_url = f"https://polygonzkevm-mainnet.g.alchemy.com/v2/{WEB3_ALCHEMY_POLYGON_ZKEVM}"
+        elif rpc_url == "" and network_name == OPTIMISM_NETWORK:
+            rpc_url = f"https://opt-mainnet.g.alchemy.com/v2/{WEB3_ALCHEMY_OPTIMISM}"
+        elif rpc_url == "" and network_name == COINBASE_BASE_NETWORK:
+            rpc_url = f"https://base-mainnet.g.alchemy.com/v2/{WEB3_ALCHEMY_BASE}"
+        elif rpc_url == "" and network_name == ARBITRUM_ONE_NETWORK:
+            rpc_url = f"https://arb-mainnet.g.alchemy.com/v2/{WEB3_ALCHEMY_ARBITRUM}"
+        elif rpc_url == "":
+            raise ValueError("Invalid network name or missing rpc_url")
+        return rpc_url
```
barakman, line 117: This function seems wrong in several different aspects:
1. Why does it need to take `rpc_url: str = None` as input to begin with?
2. Why does it check `not rpc_url` in some cases and `rpc_url == ""` in other cases?
3. Why does it check `rpc_url == ""` on every case? Surely there can be a simpler way
4. No final `else` case, is that why this function takes `rpc_url` as input to begin with?

It feels that there has to be a much less cumbersome way of writing this function, with the primary advantage being that it becomes less prone to unexpected behavior.

First off - no need to pass `rpc_url` just in order to return it as output if it contains a real value (i.e., neither `None` nor `""`).
Wherever this function is called, instead of passing that real `rpc_url` value, the caller can just use it instead of calling the function.
Subsequently everything becomes much cleaner in this function.
You can simply replace every `elif` statement with an `if` statement, and return the result directly from within each such statement.
And at the end of the function, in order to avoid unexpected behavior, you should raise an exception.

fastlane_bot/events/utils.py
```diff
@@ -591,14 +593,17 @@ def get_config(
...
     Returns
     -------
     Config
         The config object.
 
     """
     default_min_profit_gas_token = Decimal(default_min_profit_gas_token)
-
+    if rpc_url:
+        rpc_url = NetworkBase.get_rpc_url(blockchain, tenderly_fork_id, rpc_url)
```
barakman, line 606: Pointless, because at present, if `rpc_url` evaluates to a true result (i.e., it is a string of length larger than zero), then `get_rpc_url` simply returns this value as is. If anything, it should be called `if not rpc_url`.

barakman (changes requested):

fastlane_bot/config/connect.py
mikewcasale: Yes good catch. I actually think most of this logic was unnecessary. Now removed

fastlane_bot/events/utils.py
mikewcasale: see revised

fastlane_bot/config/connect.py
mikewcasale: see revised

barakman (changes requested):

barakman (approved):

fastlane_bot/events/utils.py
```diff
@@ -617,6 +621,13 @@ def get_config(
             self_fund=self_fund,
         )
         cfg.logger.info("[events.utils.get_config] Using mainnet config")
+
+    if rpc_url:
+        cfg.w3 = Web3(Web3.HTTPProvider(rpc_url, request_kwargs={"timeout": 60}))
+        cfg.w3_async = AsyncWeb3(AsyncWeb3.AsyncHTTPProvider(rpc_url))
+        if 'tenderly' in rpc_url:
+            cfg.NETWORK = cfg.NETWORK_TENDERLY
```
barakman, line 629: This looks suspicious, with respect to:
```
network = self.NETWORK if "tenderly" not in self.NETWORK else "ethereum"
```
In file `network.py`.
Can you please verify that there is no "collision" between these two?

barakman (changes requested):

fastlane_bot/events/utils.py
mikewcasale: I can confirm there is no collision between the two, however your suspicions are correctly identifying a somewhat 'hacky' solution. Unfortunately the convoluted state of the config/* folder dictates my approach here, and I suggest we address the underlying issue separately ASAP. For the current purpose however, I can confirm there is no collision.

barakman (approved):

PR #363, opened 2024-02-08
Update Base static data & small fix in Terraformer

Lesigh-3100:
-Updated static data on Base
-Fixed issue in the Terraformer that could result in losing 1 block-chunk if a number was rounded down.


run_blockchain_terraformer.py
```diff
@@ -806,7 +806,7 @@ def get_solidly_pool_creation_events_v2(
     """
     current_block = web3.eth.block_number
     events = []
-    for idx in range(int((current_block - block_number) / block_chunk_size)):
+    for idx in range(int(math.ceil((current_block - block_number) / block_chunk_size))):
         from_block = block_number + idx * block_chunk_size
         to_block = (
             from_block + block_chunk_size
```
barakman, line 812: Several notes:

1: The cast to `int` is no longer needed, since function `math.ceil` already ensures that.

2: Most of the code in these 3 functions is heavily duplicated; specifically, this part:
```
    current_block = web3.eth.block_number
    events = []
    for idx in range(int((current_block - block_number) / block_chunk_size)):
    for idx in range(int(math.ceil((current_block - block_number) / block_chunk_size))):
        from_block = block_number + idx * block_chunk_size
        to_block = (
            from_block + block_chunk_size
            if from_block + block_chunk_size < current_block
            else current_block
        )
```

3: The entire calculation above can be as simple as:
```
indexes = range(block_number, web3.eth.block_number + 1, block_chunk_size)
for from_block, to_block in zip(indexes, indexes[1:]):
    # do whatever you need to do with `from_block` and `to_block`
```

4: The code in the 3rd function is inefficient, because it checks the value of `exchange` on every iteration, even though this value remains constant throughout the loop:
```
if exchange in ["aerodrome_v2",]:
    ...
elif exchange in ["velocimeter_v2", "scale_v2"]:
    ...
```

5: Essentially, there are only 2 functions which need to be implemented here:
1. A function which retrieves `PairCreated` events, which should be called for uni_v2, velocimeter_v2 and scale_v2
2. A function which retrieves `PoolCreated` events, which should be called for uni_v3 and aerodrome_v2

Each one of these function can then be called based on the exchange in context (rather than passing that value as input).

---

To summarize all of the above, this whole section can be reduced to:
```
def _get_creation_events(block_number: int, web3: Web3, block_chunk_size, event: any) -> List:
    indexes = range(block_number, web3.eth.block_number + 1, block_chunk_size)
    return [event.get_logs(fromBlock=fromBlock, toBlock=toBlock) for fromBlock, toBlock in zip(indexes, indexes[1:])]

def get_pair_creation_events(factory_contract, block_number: int, web3: Web3, block_chunk_size=50000) -> List:
    return _get_creation_events(block_number, web3, block_chunk_size, factory_contract.events.PairCreated)

def get_pool_creation_events(factory_contract, block_number: int, web3: Web3, block_chunk_size=50000) -> List:
    return _get_creation_events(block_number, web3, block_chunk_size, factory_contract.events.PoolCreated)
```
Then subsequently:
- Function `get_pair_creation_events` events should be called for uni_v2, velocimeter_v2 and scale_v2
- Function `get_pool_creation_events` events should be called for uni_v3 and aerodrome_v2

barakman (changes requested):

barakman (approved):

PR #364, opened 2024-02-08
Simplify and improve event-fetching

barakman:
1. Simplify the code for fetching events.
3. Avoid fetching duplicated events.


zavelevsky (approved):
LGTM but please run and compare the results

mikewcasale (changes requested):
It seems like the outputs for all of the removed functions could be tested against the new function(s), in order to ensure equivalent outputs. Please provide such tests or explain your reasoning for not providing them.

barakman:
> It seems like the outputs for all of the removed functions could be tested against the new function(s), in order to ensure equivalent outputs. Please provide such tests or explain your reasoning for not providing them.

I've done it locally. Adding a test for it means preserving old code, which doesn't make much sense.

mikewcasale (approved):

PR #373, opened 2024-02-14
Fix handling of Solidly Factory ABI for equalizer & scale

Lesigh-3100:
Fixed handling of factory contract for scale & equalizer


barakman (changes requested):
In each one of the following functions:
- `get_abi`
- `get_factory_abi`
- `get_fee`

Split the handling of `self.exchange_name` into the following 6 cases:
- if `"velocimeter_v2"`...
- elif `"scale_v2"`...
- elif `"equalizer_v2"`...
- elif `"solidly_v2"`...
- elif `"aerodrome_v2"`...
- else: raise an exception

And handle each case completely separately **even** if it happens to be identical to one of the other cases.

Please use `==` only (no "if in list" or similar).

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,11 +41,21 @@ def add_pool(self, pool: Pool):
         self.pools[pool.state["address"]] = pool
 
     def get_abi(self):
-        return VELOCIMETER_V2_POOL_ABI if "velocimeter" in self.exchange_name else SOLIDLY_V2_POOL_ABI
+        if self.exchange_name == "velocimeter_v2":
+            return VELOCIMETER_V2_POOL_ABI
+        elif self.exchange_name in ["equalizer_v2"]:
```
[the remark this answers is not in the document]
Lesigh-3100: Updated

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -11,10 +11,18 @@
...
 from fastlane_bot.events.pools.base import Pool
 
+FEE_RESOLUTION_MAP = {
+    "velocimeter_v2": 10000,
+    "equalizer_v2": 10000,
+    "aerodrome_v2": 10000,
+    "velodrome_v2": 10000,
+    "scale_v2": 1000000000000000000,
+    "solidly_v2": 10000,
+}
```
barakman, line 25: Please use `1e18`, `1e5`, etc, so it is easier to spot typing mistakes.
Perhaps even better - you can use `18`, `5`, etc, and in the code itself - `10 ** FEE_RESOLUTION_MAP[x]`.

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,31 +49,70 @@ def add_pool(self, pool: Pool):
...
+        elif self.exchange_name == "scale_v2":
+            return SOLIDLY_V2_POOL_ABI
+        elif self.exchange_name == "aerodrome_v2":
+            return SOLIDLY_V2_POOL_ABI
+        elif self.exchange_name == "velodrome_v2":
+            return SOLIDLY_V2_POOL_ABI
+        elif self.exchange_name == "solidly_v2":
+            return SOLIDLY_V2_POOL_ABI
+        else:
+            raise self.WrongExchangeException(f"[exchanges/solidly_v2 get_Abi] Exchange {self.exchange_name} not in supported exchanges.")
```
barakman, line 65: You can define `ABI_MAP` similar to `FEE_RESOLUTION_MAP`, and then turn this function into:
```
if self.exchange_name in ABI_MAP:
    return ABI_MAP[self.exchange_name]
raise self.WrongExchangeException(...)
```

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,31 +49,70 @@ def add_pool(self, pool: Pool):
...
+        elif self.exchange_name == "scale_v2":
+            return SCALE_V2_FACTORY_ABI
+        elif self.exchange_name == "aerodrome_v2":
+            return SOLIDLY_V2_FACTORY_ABI
+        elif self.exchange_name == "velodrome_v2":
+            return SOLIDLY_V2_FACTORY_ABI
+        elif self.exchange_name == "solidly_v2":
+            return SOLIDLY_V2_FACTORY_ABI
+        else:
+            raise self.WrongExchangeException(f"[exchanges/solidly_v2 get_Abi] Exchange {self.exchange_name} not in supported exchanges.")
```
barakman, line 83: Same as above.

In fact, perhaps even a single variable which encompasses all the required constant information.

Something like:
```
EXCHANGE_INFO = {
    "aerodrome_v2"  : {"decimals":  5, "factory_abi": ..., "pool_abi": ...},
    "equalizer_v2"  : {"decimals":  5, "factory_abi": ..., "pool_abi": ...},
    "scale_v2"      : {"decimals": 18, "factory_abi": ..., "pool_abi": ...},
    "solidly_v2"    : {"decimals":  5, "factory_abi": ..., "pool_abi": ...},
    "velocimeter_v2": {"decimals":  5, "factory_abi": ..., "pool_abi": ...},
    "velodrome_v2"  : {"decimals":  5, "factory_abi": ..., "pool_abi": ...},
}
```

barakman (changes requested):

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: Updated to this

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: Updated

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: Updated

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,31 +49,49 @@ def add_pool(self, pool: Pool):
         self.pools[pool.state["address"]] = pool
 
     def get_abi(self):
-        return VELOCIMETER_V2_POOL_ABI if "velocimeter" in self.exchange_name else SOLIDLY_V2_POOL_ABI
+        try:
```
barakman, line 52: Don't "try".

Use "if key in dict then do stuff, else raise exception".

A "try" can fail for some unexpected reason (in opposed to the one that you're expecting it to fail on), which will make it indistinguishable from the expected reason of a failure in this case.

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,31 +49,49 @@ def add_pool(self, pool: Pool):
...
+            return EXCHANGE_INFO.get(self.exchange_name).get("pool_abi")
+        except KeyError:
+            raise self.WrongExchangeException(f"[exchanges/solidly_v2 get_Abi] Exchange {self.exchange_name} not in supported exchanges.")
 
     @property
     def get_factory_abi(self):
-        return VELOCIMETER_V2_FACTORY_ABI if "velocimeter" in self.exchange_name else SOLIDLY_V2_FACTORY_ABI
+        try:
```
barakman, line 59: Same as above

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,31 +49,49 @@ def add_pool(self, pool: Pool):
...
+        elif self.exchange_name == "velodrome_v2":
+            is_stable = await contract.caller.stable()
+            default_fee = await self.factory_contract.caller.getFee(address, is_stable)
+        elif self.exchange_name == "solidly_v2":
+            is_stable = await contract.caller.stable()
+            default_fee = await self.factory_contract.caller.getFee(address, is_stable)
+        else:
+            raise self.WrongExchangeException(f"[exchanges/solidly_v2 get_Abi] Exchange {self.exchange_name} not in supported exchanges.")
+
+        fee_float = float(default_fee) / 10 ** EXCHANGE_INFO.get(self.exchange_name).get("decimals")
```
barakman, line 86: In order to use `EXCHANGE_INFO` here in the same way as above, implement the following two global functions:
```
async def getFee1(address: str, contract: Contract, factory_contract: Contract):
    return await factory_contract.caller.getFee(address)

async def getFee2(address: str, contract: Contract, factory_contract: Contract):
    return await factory_contract.caller. getRealFee(address)

async def getFee3(address: str, contract: Contract, factory_contract: Contract):
    return await self.factory_contract.caller.getFee(address, await contract.caller.stable())
```
Then add a `"get_fee"` key in each one of the values on the `EXCHANGE_INFO` dictionary, with the value of that key being the appropriate function out of these 3 functions above.

barakman (changes requested):

fastlane_bot/events/exchanges/solidly_v2.py
mikewcasale: > Don't "try".
> 
> Use "if key in dict then do stuff, else raise exception".
> 
> A "try" can fail for some unexpected reason (in opposed to the one that you're expecting it to fail on), which will make it indistinguishable from the expected reason of a failure in this case.

The approach of using "if-then-else" logic to check for key presence in a dictionary before proceeding, as opposed to employing "try-except" for control flow and error handling, has its own trade-offs. While it addresses the concern of catching unexpected exceptions and makes the failure reasons more explicit, it also has potential downsides compared to the original "try-except" approach:

1. **Performance Considerations:** In scenarios where the expected path is the absence of errors (i.e., the key exists in the dictionary most of the time), the "try-except" pattern can be more efficient. This is because accessing a key directly and catching an exception if it doesn't exist avoids the overhead of checking the condition each time. Python's exception handling is very efficient when exceptions are rare.

2. **Readability and Pythonic Style:** Python's philosophy encourages using exceptions for control flow in certain scenarios, following the EAFP (Easier to Ask for Forgiveness than Permission) principle. This principle suggests that attempting an operation directly and handling exceptions if they occur can lead to cleaner and more readable code than pre-checking conditions. The original approach aligns more with this Pythonic style, making it more idiomatic to Python developers.

3. **Complexity in Handling Multiple Keys:** If you need to access multiple keys or perform several operations that might fail, a "try-except" block can handle all these cases concisely without multiple nested "if" checks. This can keep the code less nested and more straightforward.

4. **Error Specificity:** While the "if-then-else" approach provides clarity on why an operation failed (by checking conditions explicitly), a well-designed "try-except" block can also capture specific exceptions and provide detailed error messages. The distinction becomes less about the ability to provide error specificity and more about the stylistic and performance choices in handling expected vs. unexpected conditions.

barakman: Performance is negligible (here in particularly).
But TBH, since the `else` case (or the `except` case if using `try`) is used here just for raising an exception, it would probably be cleanest to simply do the desired operation.
In other words - no need for neither if/else nor try/except.
Simply access the dictionary, and any exception thrown as a result of that should be easy to analyze.

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: Updated

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: Updated

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,31 +60,23 @@ def add_pool(self, pool: Pool):
         self.pools[pool.state["address"]] = pool
 
     def get_abi(self):
-        return VELOCIMETER_V2_POOL_ABI if "velocimeter" in self.exchange_name else SOLIDLY_V2_POOL_ABI
+        return EXCHANGE_INFO.get(self.exchange_name).get("pool_abi")
```
barakman, line 63: You can use `[]` instead of `get`, it is cleaner and more readable.

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,31 +60,23 @@ def add_pool(self, pool: Pool):
...
 
     @property
     def get_factory_abi(self):
-        return VELOCIMETER_V2_FACTORY_ABI if "velocimeter" in self.exchange_name else SOLIDLY_V2_FACTORY_ABI
+        return EXCHANGE_INFO.get(self.exchange_name).get("factory_abi")
```
barakman, line 67: You can use `[]` instead of `get`, it is cleaner and more readable.

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -41,31 +60,23 @@ def add_pool(self, pool: Pool):
...
-
-        elif self.exchange_name in ["scale_v2", "equalizer_v2"]:
-            default_fee = await self.factory_contract.caller.getRealFee(address)
-            default_fee = float(default_fee) / 10 ** 18 if self.exchange_name == "scale_v2" else float(default_fee) / 10000
-        else:
-            is_stable = await contract.caller.stable()
-            default_fee = await self.factory_contract.caller.getFee(address, is_stable)
-            default_fee = float(default_fee) / 10000
-        return str(default_fee), default_fee
+        fee = await EXCHANGE_INFO.get(self.exchange_name).get("fee_function")(address, contract, self.factory_contract)
```
barakman, line 73: You can use `[]` instead of `get`, it is cleaner and more readable.
Also, instead of reading `EXCHANGE_INFO[self.exchange_name]` twice, you can do something like:
```
exchange_info = EXCHANGE_INFO[self.exchange_name]
fee = await exchange_info["fee_function"](address, contract, self.factory_contract)
fee_float = float(fee) / 10 ** exchange_info["decimals"]
```

barakman (changes requested):

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: updated

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: updated

fastlane_bot/events/exchanges/solidly_v2.py
Lesigh-3100: updated

barakman (approved):

PR #378, opened 2024-02-15
Adds Fantom Support

mikewcasale:
Combines issues #362 and #360 plus various configurations required for support.


fastlane_bot/helpers/routehandler.py
```diff
@@ -592,31 +605,29 @@ def _extract_single_flashloan_token(self, trade_instructions: List[TradeInstruct
...
-            flash_tokens = {
-                tknin_address :
-                {
-                "tkn": tknin_address,
-                "flash_amt": trade_instructions[0].amtin_wei,
-                "decimals": trade_instructions[0].tknin_decimals}
-            }
+        is_FL_NATIVE_permitted=False
+        if self.ConfigObj.NETWORK in [self.ConfigObj.NETWORK_ETHEREUM]:
+            is_FL_NATIVE_permitted=True
```
barakman, line 610:
```
is_FL_NATIVE_permitted = self.ConfigObj.NETWORK == self.ConfigObj.NETWORK_ETHEREUM
```

run_blockchain_terraformer.py
```diff
@@ -694,12 +712,12 @@ def organize_pool_details_solidly_v2(
...
 
-    if "velocimeter" in exchange:
+    if exchange in [VELOCIMETER_V2_NAME]:
         default_fee = factory_contract.caller.getFee(pool_address)
         default_fee = float(default_fee) / 10000
-    elif "scale" in exchange:
+    elif exchange in [SCALE_V2_NAME, EQUALIZER_V2_NAME]:
         default_fee = factory_contract.caller.getRealFee(pool_address)
-        default_fee = float(default_fee) / 10 ** 18
+        default_fee = float(default_fee) / 10 ** 18 if exchange == SCALE_V2_NAME else float(default_fee) / 10000
```
barakman, line 720: Exact same code (used to appear) in solidly_v2.py; Why is it here too?

run_blockchain_terraformer.py
```diff
@@ -1136,6 +1154,10 @@ def terraform_blockchain(network_name: str, web3: Web3 = None, start_block: int
         factory_address = row[1]["factory_address"]
         router_address = row[1]["router_address"]
         fee = row[1]["fee"]
+        is_active = row[1]["active"]
+
+        if is_active == "FALSE":
+            continue
```
barakman, line 1160: Keep it simple:
```
if row[1]["active"] == "FALSE"
```

barakman (changes requested):
File [routehandler.py](https://github.com/bancorprotocol/fastlane-bot/pull/378/files#diff-b4e47bbc00e0ca3205bce3a710016284481e51e6f4212713c899184c4975d56f) seems to contain a ton of changes, most if not all appear to be cosmetic.
Rather hard to follow each and every one of them, so please pay special attention to that.

.env.example
```diff
@@ -1,5 +1,6 @@
 export WEB3_ALCHEMY_PROJECT_ID="0-USE-YOUR-OWN-ALCHEMY-PROJECT-ID-HERE"
 export ETH_PRIVATE_KEY_BE_CAREFUL="0x123-USE-YOUR-OWN-PRIVATE-KEY-HERE"
+export WEB3_FANTOM
```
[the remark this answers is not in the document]
mikewcasale: @Lesigh-3100 ? Given that it's a free RPC, is there a key required? If so, why is `WEB3_FANTOM` formatted differently here than the others?

fastlane_bot/helpers/routehandler.py
```diff
@@ -592,31 +605,29 @@ def _extract_single_flashloan_token(self, trade_instructions: List[TradeInstruct
...
+        elif trade_instructions[0].tknin_is_native and is_FL_NATIVE_permitted:
+            tknin_address = self.ConfigObj.NATIVE_GAS_TOKEN_ADDRESS
+        elif trade_instructions[0].tknin_is_wrapped:
+            tknin_address = self.ConfigObj.WRAPPED_GAS_TOKEN_ADDRESS
         else:
-            flash_tokens = {self.wrapped_gas_token_to_native(trade_instructions[0].tknin_address): {
-                "tkn": self.wrapped_gas_token_to_native(trade_instructions[0]._tknin_address),
-                "flash_amt": trade_instructions[0].amtin_wei,
-                "decimals": trade_instructions[0].tknin_decimals}}
+            tknin_address = trade_instructions[0].tknin_address
```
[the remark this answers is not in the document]
mikewcasale: Let's leave any refactoring for a different PR. Primary objective is to get this running over the weekend to test

fastlane_bot/helpers/routehandler.py
```diff
@@ -592,31 +605,29 @@ def _extract_single_flashloan_token(self, trade_instructions: List[TradeInstruct
...
+        assert tknin_address is not None and tknin_address is not None, f"routehandler _extract_single_flashloan_token, tknin_address is {tknin_address} and tknin_address is {tknin_address}, must not be None"
```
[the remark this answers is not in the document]
mikewcasale: this is fixed now

fastlane_bot/helpers/routehandler.py
mikewcasale: The suggested change would drastically decrease readability IMO

run_blockchain_terraformer.py
mikewcasale: @Lesigh-3100 ?

run_blockchain_terraformer.py
mikewcasale: this is changed now

.env.example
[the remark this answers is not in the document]
Lesigh-3100: I formatted it differently because it's not using Alchemy. 

A key isn't required, but should be an option. For public usage, the value in .env should be "public".

run_blockchain_terraformer.py
Lesigh-3100: @barakman I pushed changes to make the code in line with the solidly_v2.py exchange style.

.env.example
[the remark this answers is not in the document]
barakman: This seems to be the only place where the `WEB3_FANTOM` environment variable is used:
```
WEB3_ALCHEMY_PROJECT_ID = os.environ.get("WEB3_FANTOM")
```
Have you verified that it achieves the desired purpose when `WEB3_FANTOM` is set to `"public"`?

run_blockchain_terraformer.py
barakman: Why duplicate the code instead of reusing it?
Ideally, this module should just call a function (or functions) in the other module.

.env.example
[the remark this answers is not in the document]
Lesigh-3100: "public" serves as the API key field - it is the correct input for the public API endpoint. Inputting it this way allows the user to change "public" with a paid API key if they desire. 

It achieves this by combining this key to the RPC_ENDPOINT in provider.py.

run_blockchain_terraformer.py
Lesigh-3100: The functions in terraformer are not async because the module does not use async to avoid rate limiting. We could consider restructuring it to use the exchange classes but it would be a major refactor.

.env.example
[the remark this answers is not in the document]
barakman: So why not use the same (or a similar) example appearing two lines above that one:
```
export WEB3_ALCHEMY_PROJECT_ID="0-USE-YOUR-OWN-ALCHEMY-PROJECT-ID-HERE"
```
?

run_blockchain_terraformer.py
barakman: You don't need to call the function as is.
There is obviously a piece of code here duplicated in two places.
You can extract that piece of code into a separate function, and then use that function where needed.

.env.example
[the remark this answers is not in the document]
Lesigh-3100: We need a distinct API key name for Fantom.

run_blockchain_terraformer.py
Lesigh-3100: I refactored this to use the solidly_v2 exchange class's functions.

.env.example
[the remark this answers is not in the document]
Lesigh-3100: I've updated the documentation for this - please let me know if you think it's clear enough now!

run_blockchain_terraformer.py
```diff
@@ -75,11 +106,27 @@
...
+
+def get_fee_2(address: str, is_stable: str, factory_contract: Contract):
+    return factory_contract.caller.getRealFee(address)
+
+
+def get_fee_3(address: str, is_stable: str, factory_contract: Contract):
+    return factory_contract.caller.getFee(address, is_stable)
+
+
+SOLIDLY_EXCHANGE_INFO = EXCHANGE_INFO
```
barakman, line 128: I don't understand where all of this (those 3 functions, as well as the "upper-case" variable which follows them) is being used.
I also don't understand the point in assigning `SOLIDLY_EXCHANGE_INFO = EXCHANGE_INFO`.
Why not just use `EXCHANGE_INFO` as is (in case it is actually used anywhere)?

barakman (changes requested):

run_blockchain_terraformer.py
barakman: Sorry for the nitpicking, but I now get the rationale of referring to it with a different name (though I'm not sure that this is indeed what you had in mind when you did so).

The cleanest solution here is probably to import it `as SOLIDLY_EXCHANGE_INFO`, and then use the latter.

Lesigh-3100: Yes I wanted to make it clear that it was exchange info only related to Soldily. That is much cleaner - I've updated it to this format.

barakman (approved):

PR #380, opened 2024-02-15
Update main.py

mikewcasale:
After testing on the current PR `fantom-rpc-flashloan-configs` branch, it was found that the argparse command-line behaviors are not backward compatible and do not work as expected. 

The current changes therefore implement backward compatibility and intuitive usage behavior.

### DIFF
--- a/fastlane_bot/bot.py
+++ b/fastlane_bot/bot.py
@@ -210,9 +210,7 @@ class CarbonBotBase:
                     f"[bot.get_curves] MUST FIX INVALID CURVE {p} [{e}]\n"
                 )
             except TypeError as e:
-                if fastlane_bot.__version__ in ["3.0.31", "3.0.32"]:
-                    pass
-                else:
+                if fastlane_bot.__version__ not in ["3.0.31", "3.0.32"]:
                     self.ConfigObj.logger.error(
                         f"[bot.get_curves] MUST FIX DECIMAL ERROR CURVE {p} [{e}]\n"
                     )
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -74,24 +74,24 @@ def process_arguments(args):
     """
     # Define the transformations for each argument
     transformations = {
-        "backdate_pools": bool,
+        "backdate_pools": lambda x: x.lower() == "true",
```
barakman, line 77: If you use `lambda x: f'{x}'.lower() == 'true'`, then you can keep all of the assignment below in their natural format.

In other words, you can continue using `someVar = False` instead of replacing it with `someVar = 'False'`, etc.

Regardless of whether or not you accept this suggestion, it might be better to implement something like:
```
is_true = lambda x: x.lower() == "true"
```
Or even:
```
def is_true(x):
    return x.lower() == "true"
```
And then use `is_true` instead of all of these lambdas.

barakman (approved):

fastlane_bot/bot.py
```diff
@@ -209,9 +210,12 @@ def get_curves(self) -> CPCContainer:
                     f"[bot.get_curves] MUST FIX INVALID CURVE {p} [{e}]\n"
                 )
             except TypeError as e:
-                self.ConfigObj.logger.error(
-                    f"[bot.get_curves] MUST FIX DECIMAL ERROR CURVE {p} [{e}]\n"
-                )
+                if fastlane_bot.__version__ in ["3.0.31", "3.0.32"]:
+                    pass
+                else:
```
barakman, line 215: Of simply:
```
if fastlane_bot.__version__ not in ["3.0.31", "3.0.32"]:
    ...
```

mikewcasale: this is changed now

main.py
mikewcasale: changed to `is_true` approach, thanks

PR #382, opened 2024-02-18
Update Base static-pool-data

Lesigh-3100:
Updated static data for Base


barakman (dismissed):

mikewcasale (commented):
Can you confirm that changes in each file are required? There seems to be a large number of diffs to all the static files, just want to be sure these are purposeful.

mikewcasale (approved):

Issue #403, opened 2024-03-05
Wrong fee resolution used for Solidly forks


Lesigh-3100:
## Bug Description
In the solidly_v2.py exchange script, the EXCHANGE_INFO defines the fee resolution for many exchanges as "5" when they should be "4" in the "decimals" field. 

Affected exchanges:
-velocimeter_v2
-equalizer_v2
-aerodrome_v2
-velodrome_v2

This has resulted in an incorrect fee stored in static data on Base for Velocimeter & Aerodrome pools. Note that static pool data on Fantom has the correct fee. 

This fix will include the fixed fee resolution & updated static data.

## Severity (High/Medium/Low)
High
## Steps to Reproduce
Run terraformer on Base & see incorrect fee assignment for affected exchanges.

## Expected Behavior
Fee resolution in the "decimals" field should be 4 for the listed exchanges. 

## Actual Behavior
Fee resolution in the "decimals" field is 5 for the listed exchanges. 

## Impact Analysis (to be filled by contributors)
- Automated Tests: Are there existing tests covering this? Do we need new tests?
No
- Risks: Potential areas affected by this bug and its probable fix
No 
- Performance: Is this bug affecting performance?
No
- Monitoring: Does this affect any current monitoring? Any new alerts needed?
No
- Logging: Any changes required in logging?
No
- Documentation: Updates needed in documentation?
No

PR #404, opened 2024-03-05
Fix fee resolution for several solidly forks

Lesigh-3100:
Description of bug:
https://github.com/bancorprotocol/fastlane-bot/issues/403


run_blockchain_terraformer.py
```diff
@@ -715,7 +715,7 @@ def organize_pool_details_solidly_v2(
 
     stable_pool = "stable" if pool_data["args"]["stable"] else "volatile"
     pool_contract = async_web3.eth.contract(address=pool_address, abi=exchange_object.get_abi())
-    fee_float = asyncio.get_event_loop().run_until_complete(asyncio.gather(exchange_object.get_fee(address=pool_address, contract=pool_contract)))
+    fee_str, fee_float = asyncio.get_event_loop().run_until_complete(asyncio.gather(exchange_object.get_fee(address=pool_address, contract=pool_contract)))[0]
```
barakman, line 718: How come this change works?
What did the call to function `run_until_complete` return before this change, and what does it return after?

barakman (changes requested):

run_blockchain_terraformer.py
Lesigh-3100: Expected return before was a single float.

After recent changes, however, it now returns List[Tuple[str, float]]

This simply takes the first and only item in the List and unpacks the tuple into str, float.

barakman (approved):

mikewcasale (approved):

Issue #412, opened 2024-03-07
Issue with self-funding mode


Lesigh-3100:
## Bug Description
Self funding mode usage had two issues that prevented it from working:
- Approval TX was written in a way that would often fail. 
- Trade amounts in TX must not be zero'd (trade everything available) for first trade. 

## Severity (High/Medium/Low)
Medium (Only related to running bot in self-funding mode)

## Steps to Reproduce
Run bot with the configuration --self_fund=True

## Expected Behavior
Bot should execute approvals and successfully close arbs with fundAndArb function. 

## Actual Behavior
Bot often failed to submit approvals and failed to generate valid transactions.

## Impact Analysis (to be filled by contributors)
- Automated Tests: Are there existing tests covering this? Do we need new tests?
No
- Risks: Potential areas affected by this bug and its probable fix
No risks expected.

- Performance: Is this bug affecting performance?
No
- Monitoring: Does this affect any current monitoring? Any new alerts needed?
No
- Logging: Any changes required in logging?
No
- Documentation: Updates needed in documentation?
No

PR #413, opened 2024-03-07
Fix self funding

Lesigh-3100:
Bug description in issue: https://github.com/bancorprotocol/fastlane-bot/issues/412


NIXBNT (approved):
Demonstrated successful self-funded strategies executed on Ethereum, Base and Fantom on Tenderly testnets. WETH approvals were executed in each

barakman (changes requested):
The error "`max fee per gas less than block base fee`" is handled independently in 3 separate places, and it is handled in a different manner in each one of these places.

Additionally, the specific logic in at least one of these places (if not in all 3 of them), looks extremely cumbersome with a bunch of try/except/if/else clauses entangled with each other.

The one specific error handled in this PR may well be just the tip of the iceberg.
As part of this PR, can you please do a more thorough work on this issue, and come up with a unified (and hopefully cleaner) way of handling these 3 scenarios?

Lesigh-3100:
> The error "`max fee per gas less than block base fee`" is handled independently in 3 separate places, and it is handled in a different manner in each one of these places.
> 
> Additionally, the specific logic in at least one of these places (if not in all 3 of them), looks extremely cumbersome with a bunch of try/except/if/else clauses entangled with each other.
> 
> The one specific error handled in this PR may well be just the tip of the iceberg. As part of this PR, can you please do a more thorough work on this issue, and come up with a unified (and hopefully cleaner) way of handling these 3 scenarios?

I've removed the duplicated handling of "`max fee per gas less than block base fee`".

It's now handled only in a single function called:
`_handle_function_build_exception`

Also, the contract functions are now built in a function called:
`build_transaction_generic`

fastlane_bot/helpers/txhelpers.py
```diff
@@ -457,47 +464,21 @@ def build_transaction_with_gas(
 
         returns: the transaction to be submitted to the blockchain
         """
+        value = src_amt if (src_address == self.ConfigObj.NATIVE_GAS_TOKEN_ADDRESS and self.ConfigObj.SELF_FUND) else None
```
barakman, line 467: 1. Please use `0` instead of `None`; it is always better to maintain the same type for a given variable, unless there really is no value of that type which is suitable for the given scenario; here, passing `value = 0` to the contract function means that it doesn't take any ETH, which is what this conditions is designated for.
2. Something doesn't feel right about the "`and`" here; for example, when the source token is ETH and self-funding is disabled, you pass `value = 0` despite the fact that an ETH amount larger than 0 seems to be needed

fastlane_bot/helpers/txhelpers.py
```diff
@@ -457,47 +464,21 @@ def build_transaction_with_gas(
...
+        transaction = self.build_transaction_generic(
+                                                    self.construct_contract_function,
+                                                    routes,
+                                                    src_amt,
+                                                    src_address,
+                                                    flashloan_struct,
+                                                    gas_price=gas_price,
+                                                    max_priority_fee=max_priority_fee,
+                                                    nonce=nonce,
+                                                    value=value
```
barakman, line 477: Minor: the indentation of all of these parameters should be reduced by a lot.

fastlane_bot/helpers/txhelpers.py
```diff
@@ -772,31 +787,20 @@ def approve_token_for_arb_contract(self, token_address: str, approval_amount: in
...
-                        f"(***2***) Error when building transaction: {e.__class__.__name__} {e}")
-            else:
-                return None
+
+        approve_tx = self.build_transaction_generic(token_contract.functions.approve,
+                                                    self.arb_contract.address,
+                                                    approval_amount,
+                                                    gas_price=current_gas_price,
+                                                    max_priority_fee=max_priority,
+                                                    nonce=self.get_nonce(),
```
barakman, line 796: Minor: the indentation of all of these parameters should be reduced by a lot.

fastlane_bot/tools/pool_shutdown.py
```diff
@@ -182,92 +182,9 @@ def generate_shutdown_tx(self, tkn: str) -> str:
             self.tx_helpers.get_max_priority_fee_per_gas_alchemy()
             * self.mgr.cfg.DEFAULT_GAS_PRICE_OFFSET
         )
-        transaction = self.build_transaction(
-            tkn=tkn,
-            gas_price=gas_price,
-            max_priority_gas=max_priority_gas,
-            nonce=self.tx_helpers.get_nonce(),
-        )
+        transaction = self.tx_helpers.build_transaction_generic(self.bancor_network_contract.functions.withdrawPOL, self.mgr.web3.to_checksum_address(tkn), gas_price=gas_price, max_priority_fee=max_priority_gas, nonce=self.tx_helpers.get_nonce())
```
barakman, line 185: Please format this into something readable.

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Updated

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Updated - please take a look at relevant changes

fastlane_bot/tools/pool_shutdown.py
Lesigh-3100: Updated this - please review & close if satisfactory.

fastlane_bot/helpers/txhelpers.py
```diff
@@ -598,7 +579,7 @@ def build_tx(
                 "from": self.wallet_address,
                 "nonce": nonce,
             }
-        if value is not None:
+        if value > 0:
```
barakman, line 582: Can remove the 'if'

barakman (changes requested):

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Is there a reason we should include the value field when it's not used? Could this cause any unexpected behaviors?

Lesigh-3100: Updated

barakman (approved):
The logical process in which a transaction is created here, is way too complicated and obfuscated, with a ton try/except and if/else clauses entangled with each other, making it impossible to figure out what's going on.
Should be fixed on a separate PR though, so approving.

Issue #419, opened 2024-03-12
Bug in events/pool/carbon_v1 parse_event


Lesigh-3100:
## Bug Description
The function `parse_event` calls function `parse_orders`, which returns a Tuple. The code below it checks if a List is returned - which should be Tuple. I will remove this handling because the check us unnecessary.

## Severity (High/Medium/Low)
Medium

## Steps to Reproduce
Run but with Carbon until parse_event is called. See an error that occurred from parsing the event. 

## Expected Behavior
Data should be parsed in the format indicated in the check `if isinstance(order0, list) and isinstance(order1, list)`.

## Actual Behavior
Data is parsed using the wrong code.

## Impact Analysis (to be filled by contributors)
- Automated Tests: Are there existing tests covering this? Do we need new tests?
No
- Risks: Potential areas affected by this bug and its probable fix
No
- Performance: Is this bug affecting performance?
No
- Monitoring: Does this affect any current monitoring? Any new alerts needed?
No
- Logging: Any changes required in logging?
No
- Documentation: Updates needed in documentation?
No

mikewcasale:
I do not believe there is a bug here - parse_orders() returns two lists (ie a tuple) and parse_event() checks if indeed these are lists. Is there some error or behavior which implies this is not working as expected? If not, I disagree that there is a bug.

Lesigh-3100:
> I do not believe there is a bug here - parse_orders() returns two lists (ie a tuple) and parse_event() checks if indeed these are lists. Is there some error or behavior which implies this is not working as expected? If not, I disagree that there is a bug.

You are right - this code works. There is an issue somewhere, however. Below is the stacktrace for where I found this exception. It -might- be related to Bancor POL, but I am not sure. 

```2024-03-12 13:11:01,979 [fastlane:INFO] - [bot._run] Found 7 eligible arb opportunities.
2024-03-12 13:11:01,979 [fastlane:INFO] - [bot.validate_pool_data] Validating pool data...
2024-03-12 13:11:03,704 [fastlane:ERROR] - [bot:run:single] tuple indices must be integers or slices, not str
2024-03-12 13:11:03,705 [fastlane:ERROR] - Error in main loop: Traceback (most recent call last):
  File "C:\Users\Kveen\PycharmProjects\fastlane-bot\main.py", line 448, in run
    handle_subsequent_iterations(
[...]

PR #426, opened 2024-03-13
Fix bug in carbon_v1 pool, function parse event 2

Lesigh-3100:
Issue: https://github.com/bancorprotocol/fastlane-bot/issues/419

This fixes a bug in `events/pools/carbon_v1.py` function `update_from_contract`.

The function calls the `strategy` function from the Carbon Controller contract.

Example return value:
```
(3743106036130323098097120681749450326077,
 '0x617d08085F68070F00fE4945c1e57013f7600b12',
 ['0x1F573D6Fb3F13d689FF844B4cE37794d79a7FF1C',
  '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'],
 [(0, 0, 67529602313984, 1961605126121825),
  (0, 0, 35462792969, 4450510153742)])
```

The problem is that the type of each order is `tuple`, and the correct handling for this format in the function `parse_event` is only applied to type `list`.

Therefore this fix adds `tuple` to the `isinstance` check in `parse_event`.

I have included a corresponding test to prove that the updated function works in `test_033_Pools.py`.

### DIFF
--- a/fastlane_bot/events/pools/carbon_v1.py
+++ b/fastlane_bot/events/pools/carbon_v1.py
@@ -79,7 +79,7 @@ class CarbonV1Pool(Pool):
         """
         order0, order1 = CarbonV1Pool.parse_orders(event_args, event_type)
         data["cid"] = event_args["args"].get("id")
-        if isinstance(order0, list) and isinstance(order1, list):
+        if isinstance(order0, (list, tuple)) and isinstance(order1, (list, tuple)):
             data["y_0"] = order0[0]
             data["z_0"] = order0[1]
             data["A_0"] = order0[2]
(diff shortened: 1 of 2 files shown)


fastlane_bot/events/pools/carbon_v1.py
```diff
@@ -79,7 +79,7 @@ def parse_event(
         """
         order0, order1 = CarbonV1Pool.parse_orders(event_args, event_type)
         data["cid"] = event_args["args"].get("id")
-        if isinstance(order0, list) and isinstance(order1, list):
+        if isinstance(order0, (list, tuple)) and isinstance(order1, (list, tuple)):
```
barakman, line 82: Get rid of this type checking, and rewrite function `parse_orders` to always return the same type.
And while you're at it, since that function isn't used anywhere else, just implement it inside this function.

barakman (changes requested):

sklbancor (approved):

PR #432, opened 2024-03-18
Update l1 gas fee logic

mikewcasale:
1.) Removed the old function
2.) Used Barak's new function (on add_mantle branch) 3.) Updated the Gas Oracle addresses for Base & Optimism in network.py, per the official documentation (optimism - https://docs.optimism.io/builders/tools/build/oracles#gas-oracle) and (base - https://docs.base.org/base-contracts/#base-mainnet) 4.) Verified new gas function is called in TxHelpers (removes intermediary step which called `_get_layer_one_gas_fee_loop`)

### DIFF
--- a/fastlane_bot/helpers/txhelpers.py
+++ b/fastlane_bot/helpers/txhelpers.py
@@ -133,8 +133,8 @@ class TxHelper:
     @property
     def deadline(self):
         return (
-            self.ConfigObj.w3.eth.getBlock("latest")["timestamp"]
-            + self.ConfigObj.DEFAULT_BLOCKTIME_DEVIATION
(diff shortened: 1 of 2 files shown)


fastlane_bot/helpers/txhelpers.py
```diff
@@ -214,17 +214,17 @@ def _get_prices_info(
             expected_profit_usd: Decimal,
             expected_profit_eth: Decimal,
             signed_arb_tx
```
Lesigh-3100, line 216: Could you change the param signed_arb_tx to raw_transaction?

I think it will be more clear as the input to this function is: signed_arb_tx.rawTransaction

Lesigh-3100 (changes requested):

PR #434, opened 2024-03-18
Update NBTest_070_TestCarbonTradeSplitter.py

Lesigh-3100:
Add test case that adds an additional Carbon exchange with ETH & WETH.

It splits 1 TradeInstruction into 4.

### DIFF
--- a/resources/NBTest/NBTest_070_TestCarbonTradeSplitter.py
+++ b/resources/NBTest/NBTest_070_TestCarbonTradeSplitter.py
@@ -69,7 +69,7 @@ class Pool:
 @dataclass
 class Config:
     ZERO_ADDRESS = NONE_ADDRESS
-    CARBON_V1_FORKS = [CARBON_V1_NAME]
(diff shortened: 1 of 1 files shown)


resources/NBTest/NBTest_070_TestCarbonTradeSplitter.py
```diff
@@ -28,6 +28,7 @@
 BANCOR_V2_NAME      = 'unique_id_1'
 CARBON_V1_NAME      = 'unique_id_2'
 PANCAKESWAP_V2_NAME = 'unique_id_3'
+ALTERED_CARBON = 'unique_id_4'
```
barakman, line 31: Semantics: append `_NAME` in order to make it aligned with the others.

barakman (changes requested):
The ID of the new Carbon fork should be added as one of the Carbon V1 forks in the Config object.

As it stands, you are testing a single-deployment scenario:
```
class Config:
    ...
    CARBON_V1_FORKS = [CARBON_V1_NAME]
    ...
```

Issue #415, opened 2024-03-08
ABI Cleanup


barakman:
## Enhancement Description

1. Rewrite every ABI as a dictionary declared in a properly-formatted manner, rather than as a (very long) string which is json-loaded into a dictionary during runtime
2. Keep in every ABI only the functions and events that the bot relies on
3. Merge the ABIs of all DEX contracts into a single ABI

Bullet 3 is doable only if no two functions which differ only by the return-value type exist within the different DEX contracts.
Even the potential of this possibility in the future (as more DEXs are added) might be a good reason for avoiding it altogether.
The main advantage here is that it would revoke the need to interact with different DEX contracts using different ABIs.

## Rationale

The ABI of a contract doesn't need to include all the functions and events in the contract.
It is sufficient for every ABI to include only the functions and events that are actually used in the bot.
The main advantage in the current implementation is that every ABI can be copied as is from etherscan or similar, and then pasted directly into the code without "having to think about it any further".
The main disadvantage is a ton of redundant information which needs to be maintained in the code.
And keeping it as a string makes it even more difficult, both in terms of readability and in terms of maintenance.
For example, searching through functions or events in an ABI is extremely difficult under the current implementation.

## Impact Analysis
- `"true"` will need to change to `True`
- `"false"` will need to change to `False`
- In a given ABI, any function within a group of overloaded functions (same function name, different input argument types) will need to be called using a "slightly annoying" syntax

As a example for the overloaded-functions issue, suppose a ABI with the following two functions:
1. `function getFee(uint256 id, bool stable) external returns (uint256)`
2. `function getFee(uint128 id) external returns (uint256)`

The way to call each one of these functions with `id = 5`, for example, is something like:
1. `contract.getFee["uint256,bool"](5)`
2. `contract.getFee["uint128"](5)`

barakman:
Supplemental - parameter names in an ABI are generally optional:

Function parameter names have no impact whatsoever, so they are only "nice to have".

Event parameter names impact how the client's side web3 package provides the event to the caller.
They do not need to be identical to the ones in the contract, which means that the programmers on the client side can rename them as desired.
This becomes useful when several events on different ABIs are identical with the exception of parameter names.
By aligning these names under a single convention, one can simplify the handling of the related events.

[...]

PR #435, opened 2024-03-19
ABI Cleanup

barakman:
Handles issue #415 partially:
1. Rewrite every ABI as a dictionary declared in a properly-formatted manner, rather than as a (very long) string which is json-loaded into a dictionary during runtime
2. Keep in every ABI only the functions and events that the bot relies on


mikewcasale (dismissed):

sklbancor (dismissed):
overall this seems like a trivial change; assuming tests pass [and fingers crossed that our tests cover all ABI use cases and we are not missing some ABI here] I don't see any reasons to not merge it

barakman:
> overall this seems like a trivial change; assuming tests pass [and fingers crossed that our tests cover all ABI use cases and we are not missing some ABI here] I don't see any reasons to not merge it

Trivial though pretty difficult to complete.
I've actually implemented a couple of scripts in order to achieve that:
1. A script which re-parses this file and changes its contents from JSON strings into JSON objects
2. A script which scans the entire code-base, finds all functions and events which are potentially being used, and removes all the rest

sklbancor:
> > overall this seems like a trivial change; assuming tests pass [and fingers crossed that our tests cover all ABI use cases and we are not missing some ABI here] I don't see any reasons to not merge it
> 
> Trivial though pretty difficult to complete. I've actually implemented a couple of scripts in order to achieve that:
> 
> 1. A script which re-parses this file and changes its contents from JSON strings into JSON objects
> 2. A script which scans the entire code-base, finds all functions and events which are potentially being used, and removes all the rest from those JSON objects

Jupyter actually allows you to do No 1 -- albeit possibly not in the size required. But with the snippet

```
data = json.loads(data_s)
data
```
you produce output that then you can import into Python

barakman:
> > > overall this seems like a trivial change; assuming tests pass [and fingers crossed that our tests cover all ABI use cases and we are not missing some ABI here] I don't see any reasons to not merge it
> > 
> > 
> > Trivial though pretty difficult to complete. I've actually implemented a couple of scripts in order to achieve that:
> > 
> > 1. A script which re-parses this file and changes its contents from JSON strings into JSON objects
> > 2. A script which scans the entire code-base, finds all functions and events which are potentially being used, and removes all the rest from those JSON objects
> 
> Jupyter actually allows you to do No 1 -- albeit possibly not in the size required. But with the snippet
> 
> ```
> data = json.loads(data_s)
> data
> ```
> 
> you produce output that then you can import into Python

That's pure Python, no need for Jupyter.

But that's not as simple, since I wanted to format the primary fields with a standard indentation of 4 spaces, and the internal fields (the values of 'input' and 'output') in a single line.

A kind of a mitigation between a readable file and a relatively small file, since applying the indentation for everything would have resulted with a huge file, and not applying it at all would have resulted with an unreadable file.

Also, I wanted to rearrange the fields of each element, starting with type ("function" or "event"), then name, and so on, since the Solidity compiler arranges them in a pretty weird order IMO.

So in short, a script was required hee as well.

sklbancor (approved):

fastlane_bot/tests/test_037_Exchanges.py
```diff
@@ -174,7 +173,7 @@ def test_test_uniswap_v2_exchange_fork():
     async def test_uniswap_v2_exchange_fork():
```
sklbancor, line 173: Unrelated to this PR but I need this comment to make it another issue

```
async def test_uniswap_v2_exchange_fork():
        assert (pancake_v2_exchange.exchange_name in "pancakeswap_v2"), f"Wrong exchange name. Expected pancakeswap_v2, got {pancake_v2_exchange.exchange_name}"
        assert (pancake_v2_exchange.base_exchange_name in "uniswap_v2"), f"Wrong base exchange name. Expected uniswap_v2, got {pancake_v2_exchange.base_exchange_name}"  
```

The whole point of defining constants is that they abstract from the underlying values. The only reason why we define

`PANCAKESWAP_V2_NAME = "pancakeswap_v2"`

is because this makes debugging easier. However, if at one point we would like to change 

`PANCAKESWAP_V2_NAME = 1234`

then the code should still work as previously.

Relatedly -- tests are also contracts with the users of the library. If you do 

`assert PANCAKESWAP_V2_NAME == "pancakeswap_v2"`

then this means that the user of the library can rely on this association being true (and in SemVer this means that if ever you can it you need to change the major version number)

sklbancor (approved):

PR #436, opened 2024-03-19
Add handling for specific Uni V3 exchanges on Mantle

Lesigh-3100:
These exchanges on Mantle use Uniswap V3 Router 01:
-butter_v3
-agni_v3

This change adds the exchanges to the list of exchanges that should use Router 01 in the `handle_custom_data_extras` function in routehandler.py.

### DIFF
--- a/fastlane_bot/config/constants.py
+++ b/fastlane_bot/config/constants.py
@@ -11,3 +11,9 @@ FLASHLOAN_FEE_MAP = {
     "fantom": 0.0003,
     "mantle": 0,
 }
+
+ETHEREUM = "ethereum"
+PANCAKESWAP_V3_NAME = "pancakeswap_v3"
+BUTTER_V3_NAME = "butter_v3"
+AGNI_V3_NAME = "agni_v3"
+CLEOPATRA_V3_NAME = "cleopatra_v3"
--- a/fastlane_bot/config/network.py
+++ b/fastlane_bot/config/network.py
@@ -216,9 +216,6 @@ class ConfigNetwork(ConfigBase):
     SOLIDLY_V2_NAME = "solidly_v2"
     VELODROME_V2_NAME = "velodrome_v2"
     SHIBA_V2_NAME = "shiba_v2"
-    BUTTER_V3_NAME = "butter_v3"
-    AGNI_V3_NAME = "agni_v3"
-    CLEOPATRA_V3_NAME = "cleopatra_v3"
     # Base Exchanges
     AERODROME_V2_NAME = "aerodrome_v2"
     AERODROME_V3_NAME = "aerodrome_v3"
--- a/fastlane_bot/config/network.py
+++ b/fastlane_bot/config/network.py
@@ -218,7 +218,7 @@ class ConfigNetwork(ConfigBase):
     SHIBA_V2_NAME = "shiba_v2"
     BUTTER_V3_NAME = "butter_v3"
     AGNI_V3_NAME = "agni_v3"
-
+    CLEOPATRA_V3_NAME = "cleopatra_v3"
     # Base Exchanges
     AERODROME_V2_NAME = "aerodrome_v2"
     AERODROME_V3_NAME = "aerodrome_v3"
--- a/fastlane_bot/config/network.py
+++ b/fastlane_bot/config/network.py
@@ -216,6 +216,8 @@ class ConfigNetwork(ConfigBase):
     SOLIDLY_V2_NAME = "solidly_v2"
     VELODROME_V2_NAME = "velodrome_v2"
     SHIBA_V2_NAME = "shiba_v2"
+    BUTTER_V3_NAME = "butter_v3"
+    AGNI_V3_NAME = "agni_v3"
 
     # Base Exchanges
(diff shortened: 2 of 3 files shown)


fastlane_bot/helpers/routehandler.py
```diff
@@ -320,7 +320,7 @@ def handle_custom_data_extras(self, platform_id: int, custom_data: bytes, exchan
 
         assert custom_data in "0x", f"[routehandler.py handle_uni_v3_router_switch] Expected the custom data field to contain '0x', but it contained {custom_data}. This function may need to be updated."
         if platform_id == self.ConfigObj.network.EXCHANGE_IDS.get(self.ConfigObj.network.UNISWAP_V3_NAME):
-            if self.ConfigObj.network.NETWORK in "ethereum" or exchange_name in self.ConfigObj.network.PANCAKESWAP_V3_NAME:
+            if self.ConfigObj.network.NETWORK in "ethereum" or exchange_name in [self.ConfigObj.network.PANCAKESWAP_V3_NAME, self.ConfigObj.network.BUTTER_V3_NAME, self.ConfigObj.network.AGNI_V3_NAME, self.ConfigObj.CLEOPATRA_V3_NAME]:
```
barakman, line 323: The logic in this function is really obfuscated for no reason.

For one, the first `if` can generally be removed altogether.

Another thing is that the assertion statement is baffling for several different reasons, such as:
- The condition being tested is whether or not the string `"0x"` contains the value of `custom_data`, which doesn't make any sense, and is probably only true because the value of `custom_data` itself is `"0x"` (assuming that it is never `"0"` neither `"x"`)
- The error-message says the exact opposite of that condition, namely, that the value of `custom_data` should contain the string `"0x"`

Most likely, there is no reason to assert anything other than `custom_data == "0x"`, which should be done only when the input value of `custom_data` is overridden in each one of the `if` statements which follow that assertion.

In short, the code in this function should be something like:
```
if <the condition in the first 'if' statement>:
    assert custom_data == "0x"
    return '0x0000000000000000000000000000000000000000000000000000000000000000'
if <the condition in the second 'if' statement>:
    assert custom_data == "0x"
    return '0x0100000000000000000000000000000000000000000000000000000000000000'
return custom_data
```
I think that adding more logic instead of simplifying it is counterproductive.

barakman (changes requested):

fastlane_bot/config/network.py
```diff
@@ -216,7 +216,9 @@ class ConfigNetwork(ConfigBase):
     SOLIDLY_V2_NAME = "solidly_v2"
     VELODROME_V2_NAME = "velodrome_v2"
     SHIBA_V2_NAME = "shiba_v2"
-
+    BUTTER_V3_NAME = "butter_v3"
```
mikewcasale, line 219: I would really like us to stop adding info to the config object(s). Instead, can you move these to the constants.py file? In the *very* near future I'm hoping to dissolve all config objects, so we can make it easier if we stop adding to it asap.

fastlane_bot/helpers/routehandler.py
mikewcasale, line 323: I would suggest changing the conditional logic to `x=="ethereum"` instead of `x in "ethereum"`. On a related note, it would be better to define "ethereum" in constants.py, however it may not prudent to change this here and not everywhere else... so for the time being it's OK but just want to call it out for future reference.

mikewcasale (changes requested):

fastlane_bot/helpers/routehandler.py
Lesigh-3100: I completely agree with your assessment & have opened an issue to address this upstream so that the logic here can be more straightforward: https://github.com/bancorprotocol/fastlane-bot/issues/440

However, in the interest of time, does this need to be refactored immediately?

Lesigh-3100: Updated this - could you please look again?

fastlane_bot/config/network.py
Lesigh-3100: Updated this - could you please look again?

fastlane_bot/helpers/routehandler.py
barakman: No, I think that it is 99.99% safe; so given the linked issue, resolving.

fastlane_bot/config/constants.py
```diff
@@ -11,3 +11,9 @@
     "fantom": 0.0003,
     "mantle": 0,
 }
+
+ETHEREUM = "ethereum"
+PANCAKESWAP_V3_NAME = "pancakeswap_v3"
+BUTTER_V3_NAME = "butter_v3"
+AGNI_V3_NAME = "agni_v3"
+CLEOPATRA_V3_NAME = "cleopatra_v3"
```
barakman, line 19: It's already in network.py, why here too?

barakman (changes requested):

fastlane_bot/config/constants.py
Lesigh-3100: @mikewcasale above asked to stop adding to network.py

barakman (approved):

PR #451, opened 2024-03-20
Add strategy_id to terraformer

Lesigh-3100:
The PR that this branch is based on introduced the "strategy_id" field into the static data (and Pool class) data structure. This field is only applicable for Carbon pools, and is set to 0 for all other pools to maintain consistent typing (int). This change to the Terraformer adds the missing field when pool data is generated.

I have added a corresponding test that ensures that this field is set for the four main data processing functions in the Terraformer. These tests can be expanded further into a test suite for the Terraformer, however, that is out of the scope of this PR.

### DIFF
--- a/fastlane_bot/tests/test_072_TestTerraformer.py
+++ b/fastlane_bot/tests/test_072_TestTerraformer.py
@@ -1,0 +1,123 @@
+
+'''
+This module tests the Terraformer
+'''
+import pytest
+import asyncio
+
+from unittest.mock import MagicMock, AsyncMock
+
+from json import dumps
+from pytest import fixture
+from dataclasses import dataclass
+from run_blockchain_terraformer import organize_pool_details_balancer, organize_pool_details_uni_v2, organize_pool_details_solidly_v2, organize_pool_details_uni_v3
+
+
+def test_organize_pool_details_uni_v2_strategy_id():
+    # Mock the input parameters
+    pool_data = {
+        "args": {
+            "pair": "0x123",
+            "token0": "0x456",
+            "token1": "0x789"
+        },
+        "blockNumber": 123456
+    }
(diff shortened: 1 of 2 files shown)


barakman (changes requested):
Please explain what this change addresses and/or fix.
Please add a corresponding test of some sort.

Lesigh-3100:
> Please explain what this change addresses and/or fix. Please add a corresponding test of some sort.
I've updated the explanation of this PR & added tests.

barakman (changes requested):
Tests are failing with:
```
TypeError: get_token_and_fee() got an unexpected keyword argument 'exchange_name'
```
Which is pretty weird, since this function does take an input argument by that name.
But it nevertheless requires looking into it.

fastlane_bot/tests/test_072_TestTerraformer.py
```diff
@@ -0,0 +1,123 @@
...
+'''
+import pytest
+import asyncio
+
+from unittest.mock import MagicMock, AsyncMock
+
+from json import dumps
+from pytest import fixture
+from dataclasses import dataclass
+from run_blockchain_terraformer import organize_pool_details_balancer, organize_pool_details_uni_v2, organize_pool_details_solidly_v2, organize_pool_details_uni_v3
```
barakman, line 13: The following modules aren't being used; please remove the corresponding `import` statements:
```
import pytest
import asyncio
from json import dumps
from pytest import fixture
from dataclasses import dataclass
```

barakman (approved):

fastlane_bot/tests/test_072_TestTerraformer.py
Lesigh-3100: Done - thank you.

PR #452, opened 2024-03-20
Add missing import to txhelpers

Lesigh-3100:
Calling `asyncio.gather` in `TxHelpers` fails due to missing import of `nest_asyncio` & missing `nest_asyncio.apply()`

### DIFF
--- a/fastlane_bot/helpers/txhelpers.py
+++ b/fastlane_bot/helpers/txhelpers.py
@@ -8,6 +8,7 @@ __VERSION__ = "1.0"
 __DATE__ = "01/May/2023"
 
 import asyncio
+import nest_asyncio
 from _decimal import Decimal
 
 # import itertools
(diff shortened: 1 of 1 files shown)


barakman (approved):
Well the PR title and description are wrong, since the problem is a missing function-call, not a missing import-statement, but assuming that you've verified this change (i.e., that it actually resolves an existing problem somewhere) - approving.

barakman:
> > Well the title and description of this PR are wrong, since the problem is a missing function call, not a missing import statement, but assuming that you've verified this change (i.e., that it actually resolves an existing problem somewhere) - approving.
> 
> I verified that these changes fixed the issue. It was not something that could have been found by current testing infrastructure.

So you need to add a test which proves the necessity of this change (i.e., fails without it and passes with it).

PR #478, opened 2024-03-22
Simplify class TxHelpers

barakman:
Some of the changes in the `TxHelpers` class:

1. Removed the `DEFAULT_GAS_PRICE_OFFSET` ("increase by 9%") factor applied on the max priority fee:
   - It is applied only in chains where the bot uses EIP1559 (Ethereum and Base), and the bot shouldn't rely on this partial solution, even if that solution appears to be working somehow.
   - It reduces the expected execution time of a transaction, making it more likely to succeed while the arbitrage opportunity is still viable. But on the same time, it makes that arbitrage opportunity less likely to be viable in the first place (and it makes it less likely to be viable, more or less by the same magnitude of 9%). So there's no clear advantage in this heuristics to begin with, and this is regardless of that specific 9% default value.

2. Removed the parsing of the "baseFee" error-message returned from the node. This error happens when the block's `baseFee` becomes higher than the transaction's `maxGasFee`, which theoretically yields a negative tip to the miner, hence the transaction is immediately reverted. Recall that the miner gets paid `maxGasFee - baseFee`, regardless of the specified `maxPriorityFee`, which indicates the maximum tip and not the actual tip. But even before this happens, the block's `baseFee` may become sufficiently high to bring the effective tip paid to the miner extremely small, thus leading the transaction to linger for a very long time, and most likely become invalid (arbitrage already fulfilled) by the time it gets executed. So the only difference between these two scenarios is a reverted transaction vs a stuck transaction. Both can sporadically happen, just like front-running can sporadically invalidate arbs submitted by this bot, for example. This error can be handled in a "natural manner", at the next iteration of the bot's infinite loop, hence there is no need to tailor-make a special "handle error and retry" method for this specific error.

3. Fetching the `maxPriorityFee` via web3 instead of by sending an HTTP request to the node. It shouldn't yield a notable impact, but, why invent the wheel?

4. Fetching the `gasPrice` instead of the `baseFee`, since the latter requires fetching the entire block, which most likely takes longer to complete. In the case of chains which already support EIP1559 (e.g., Ethereum, starting from the London fork), this value indicates the sum of the current block's `baseFee` and the average `maxPriorityFee`, hence it can be used as is when setting the transaction's `maxGasFee`. And of course, in the case of chains which do not yet support EIP1559, it can be used as is in order to set the transaction's `gasPrice`.

5. After the transaction is built, it already includes the estimated gas, so no need to then send another `estimateGas` request to the node.

6. Fetching the access-list via web3 instead of by sending an HTTP request to the node, because - again - why invent the wheel?

7. After the access-list is added to the transaction, it already includes the new estimated gas, so no need to then send another `estimateGas` request to the node.

8. Removed the `EXPECTED_GAS_MODIFIER` ("decrease to 85%") factor used for assessing the actual gas cost of the transaction.

9. Replaced fetching the latest block (used in order to specify max block number when sending a private transaction to Alchemy), with fetching only the block number itself, which is probably (or at least hopefully) more efficient.

10. Removed support for legacy-transactions; all chains are assumed to be supporting EIP-1559.

11. Sending `eth_sendPrivateTransaction` HTTP requests via web3 instead of via the node. This change includes the removal of the `retries=3` parameter, which needs to be verified.

12. Instead of the `DEFAULT_GAS_PRICE_OFFSET` ("increase by 9%") factor applied on the max priority fee, introduced a `GAS_STRATEGY` hook function with which the user can configure their desired gas-fee strategy.

13. Ensured that access-list is applied on Ethereum transactions; it is not applied in the current version due to a bug in the code, and even after this PR, it will not be tested within Tenderly (when the target chain is Ethereum) because of that same bug; see issue #519.

14. Constructing the `tx` dictionary manually and then setting its `gas` attribute by calling `estimate_gas`, instead of via `build_transaction` to do both of these actions; Following that, setting the `accessList` attribute of this dictionary, and possibly updating its `gas` attribute, by calling `create_access_list`; The gas-fee attributes (`maxFeePerGas` and `maxPriorityFeePerGas`) are set only at the end, in order to avoid getting the "max fee per gas less than block base fee" error.


fastlane_bot/helpers/txhelpers.py
```diff
@@ -11,29 +11,22 @@
...
 from alchemy import Network, Alchemy
 from web3.exceptions import TimeExhausted
 
-# from fastlane_bot.config import *  # TODO: PRECISE THE IMPORTS or from .. import config
-# from fastlane_bot.db.models import Token, Pool
-# import fastlane_bot.config as c
-# from fastlane_bot.tools.cpc import ConstantProductCurve
 from fastlane_bot.config import Config
-from fastlane_bot.data.abi import *  # TODO: PRECISE THE IMPORTS or from .. import abi
+from fastlane_bot.data.abi import ERC20_ABI
```
sklbancor, line 23: may I suggest `from fastlane_bot.data import abi`
and then use it as `abi.ERC20_ABI`

this avoids having to import the ABIs one by one, pollutes the namespace less, and is easier to understand without IDE

fastlane_bot/helpers/txhelpers.py
```diff
@@ -44,15 +37,13 @@ class TxHelpers:
...
-    # This is used for the Alchemy SDK
-    network = Network.ETH_MAINNET
 
     def __post_init__(self):
 
         if self.ConfigObj.network.DEFAULT_PROVIDER != "tenderly":
             self.alchemy = Alchemy(
                 api_key=self.ConfigObj.WEB3_ALCHEMY_PROJECT_ID,
-                network=self.network,
+                network=Network.ETH_MAINNET,
```
sklbancor, line 46: I appreciate that the functionality has not changed, but do we think it is a good idea to hardcode `ETH_MAINNET` in the middle of a generic class?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -66,58 +57,6 @@ def __post_init__(self):
         self.wallet_address = str(self.local_account.address)
 
         self.alchemy_api_url = self.ConfigObj.RPC_URL
-        self.nonce = self.get_nonce()
```
sklbancor, line 69: admittedly the diffs are a bit confusing; can you explain where this code went, or whether it was really deleted?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -179,16 +135,30 @@ def validate_and_submit_transaction(
         else:
             current_gas_price = arb_tx["gasPrice"]
 
-        signed_arb_tx = self.sign_transaction(arb_tx)
+        signed_arb_tx = self.web3.eth.account.sign_transaction(arb_tx, self.ConfigObj.ETH_PRIVATE_KEY_BE_CAREFUL)
```
sklbancor, line 138: what happened to self.sign_tx? why was it removed?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -455,35 +417,29 @@ def build_tx(
...
             self.wallet_address = self.ConfigObj.BINANCE14_WALLET_ADDRESS
             
-        # if "tenderly" in self.web3.provider.endpoint_uri:
-        #     print("Tenderly network detected: Manually setting maxFeePerFas and maxPriorityFeePerGas")
-        #     max_gas_price = 3
-        #     max_priority_fee = 3
-
         if self.ConfigObj.NETWORK in ["ethereum", "coinbase_base"]:
             tx_details = {
-                "type": "0x2",
```
sklbancor, line 472: why the change string vs int? are both supported?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -551,148 +507,60 @@ def _submit_transaction(self, tx_hash) -> str:
             )
             return None
 
-    def sign_transaction(self, transaction: Dict[str, Any]) -> Dict[str, Any]:
```
sklbancor, line 554: see comment above: I would probably not remove this abstraction layer here, it may have been put in for a reason and may be useful

sklbancor (changes requested):
some small stuff; does not necessarily need to be changed but I'd like to discuss

more generally -- it is hard to review the changes; it looks like functionality changed somewhat and things moved around a lot so the diffs are not particularly useful; with good test coverage this is not a problem -- but given how poor our test coverage seems to be: are you sure this does not introduce bugs?

fastlane_bot/bot.py
```diff
@@ -93,7 +92,7 @@ class CarbonBotBase:
         the database manager.
     TxRouteHandlerClass
         ditto (default: TxRouteHandler).
-    TxHelpersClass: class derived from TxHelpersBase
```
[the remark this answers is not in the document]
barakman: Keeping changes strictly related to this PR.
The "ditto" here (whatever that even means) refers to a class outside the scope of this PR.

fastlane_bot/helpers/txhelpers.py
barakman: The comment which was removed (in red) requests the exact opposite.
There's no added value in using `abi.ERC20_ABI` rather than using `ERC20_ABI` across the code.
In fact, there's a little added value in the opposite, as it saves a bit of test.
Anyways, this is minor, so resolving.

fastlane_bot/helpers/txhelpers.py
barakman: No, this is probably as wrong as it was before.
Can be dealt with in subsequent cleanup tasks (which there will probably several more of).

fastlane_bot/helpers/txhelpers.py
barakman: Clear up code.
It was a single-line function used only in two places in the code.
Placing that single line directly in each one of these 2 places, makes it slightly easier to follow the code, and also cleans it up a bit.

fastlane_bot/helpers/txhelpers.py
barakman: No abstraction here. It probably used to be a larger function, which got smaller over time.
It also appears as an internal function in a few other modules, making it a little difficult to follow the code when searching for it.
Anyways, it was a single-line function at this point, so moved that line directly to where the function was called, in order to make the code easier to read.

fastlane_bot/helpers/txhelpers.py
barakman: `2` is cleaner than `"0x2"`.
No need to use strings where not necessary.

fastlane_bot/helpers/txhelpers.py
barakman: The value of `self.nonce` is not used anywhere in this module or by any other module which interacts this module.

fastlane_bot/events/utils.py
```diff
@@ -2063,28 +2007,4 @@ def check_and_approve_tokens(tokens: List, cfg) -> bool:
     :param cfg: the config object
 
     """
-
```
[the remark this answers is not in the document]
barakman: That's the purpose of this task - cleanup.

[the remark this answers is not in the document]
sklbancor: cleanup is about either removing unused code or doing the same thing more efficiently. How confident are you that the code still operates the same way in all relevant real world scenarios?

fastlane_bot/helpers/txhelpers.py
sklbancor: then it probably should be noted down as in issue.

fastlane_bot/helpers/txhelpers.py
sklbancor: that was just the beginning of a code block of a few dozen lines that you deleted and that was doing stuff. same qn as above -- how confident are you that this code really was not needed?

fastlane_bot/helpers/txhelpers.py
sklbancor: fwiw I disagree. the original code introduced an abstraction for signing transactions which was a/ deliberate, and b/ IMO usually a good thing. we can discuss this -- but removing abstractions is not often warranted

fastlane_bot/helpers/txhelpers.py
sklbancor: disagree. it abstracts from the exact use of the web3 library. maybe this is the right thing to do, but it is not self evident

barakman:
Open questions regarding the `TxHelpers` class:

1. The value of `self.wallet_address` is set upon class initialization, but also in function `_build_transaction`, which permanently changes it to `BINANCE14_WALLET_ADDRESS`. Is this the desired behavior?

2. In function `validate_and_submit_transaction`, the pending block is used when calling `build_transaction_with_gas`, but the latest block is used when calling `submit_private_transaction`. Is this the desired behavior?

3. In function `validate_and_submit_transaction`, the multiplication by `EXPECTED_GAS_MODIFIER` is applied on the base gas cost, but not on the L1-Data part which is sometimes added to it. Is this the desired behavior?

fastlane_bot/events/utils.py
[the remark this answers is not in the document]
barakman: I'm actually pretty confident that some stuff may have been broken as a result of this PR, which we shall fix by testing it, and not by keeping the previous code or reverting back to it (due to its poor shape).
The purpose of this PR is to end up with cleaner and more robust code, which will be beneficial in the long term (and most likely also in the short term).

fastlane_bot/helpers/txhelpers.py
barakman: The code in the two internal functions - `_get_transaction_info` and `_get_prices_info` was moved into the only function which calls them (one by one) - `validate_and_submit_transaction`, which will later probably go through some additional renovation until its flow is clear and easy to follow.

fastlane_bot/helpers/txhelpers.py
barakman: Noted.

fastlane_bot/helpers/txhelpers.py
barakman: Noted.

fastlane_bot/events/utils.py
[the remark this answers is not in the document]
sklbancor: because of the poor state of our tests I am worried that this code may break something not-so-obvious in the bot, and I am not confident that we'll be able to catch it before it goes into production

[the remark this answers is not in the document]
barakman: I agree, but the same arguments holds even prior to all these changes.
The code is already broken in a bunch of non-obvious places, which occasionally emerge during runtime.
So I've opted to demolish and rebuild this part of the project.
Ultimately, the TX-Helper functions should all be fairly simple, readable and easy to implement.
As it stands, there are a bunch of culprits, including functions which return a different number of argument on different execution paths (sometimes even not returning anything when they should be), piece of code which are duplicated (or worse, partially-duplicated) across several different modules, etc.
This needs to be resolved.

sklbancor:
> Open questions regarding the `TxHelpers` class:
> 
> 1. The value of `self.wallet_address` is set upon class initialization, but also in function `_build_transaction`, which permanently changes it to `BINANCE14_WALLET_ADDRESS`. Is this the desired behavior?
> 2. In function `validate_and_submit_transaction`, the pending block is used when calling `build_transaction_with_gas`, but the latest block is used when calling `submit_private_transaction`. Is this the desired behavior?
> 3. In function `validate_and_submit_transaction`, the multiplication by `EXPECTED_GAS_MODIFIER` is applied on the base gas cost, but not on the L1-Data part which is sometimes added to it. Is this the desired behavior?

no idea, but 1/ - probably not would be my guess...

barakman:
> Open questions regarding the `TxHelpers` class:
> 
> 1. The value of `self.wallet_address` is set upon class initialization, but also in function `_build_transaction`, which permanently changes it to `BINANCE14_WALLET_ADDRESS`. Is this the desired behavior?
> 2. In function `validate_and_submit_transaction`, the pending block is used when calling `build_transaction_with_gas`, but the latest block is used when calling `submit_private_transaction`. Is this the desired behavior?
> 3. In function `validate_and_submit_transaction`, the multiplication by `EXPECTED_GAS_MODIFIER` is applied on the base gas cost, but not on the L1-Data part which is sometimes added to it. Is this the desired behavior?
> 4. In function `build_transaction_with_gas`, there's an attempt to catch a `"max fee per gas less than block base fee"` error even though this function does actually submit any transaction. What exactly is the point in that?

1. Fixed
2. Fixed to use the pending block in all cases
3. Waiting for decision
4. No need to fix; error occurs when building a transaction, not when submitting one

fastlane_bot/config/provider.py
```diff
@@ -190,12 +183,9 @@ def __init__(self, network: ConfigNetwork, **kwargs):
...
             abi=FAST_LANE_CONTRACT_ABI,
         )
-        self.ARB_CONTRACT_VERSION = self.BANCOR_ARBITRAGE_CONTRACT.caller.version()
 
-        reward_percent, max_profit = self.BANCOR_ARBITRAGE_CONTRACT.caller.rewards()
-
-        self.ARB_REWARD_PERCENTAGE = str(int(reward_percent) / 1000000)
-        self.ARB_MAX_PROFIT = str(int(max_profit) / (10**18))
+        self.ARB_CONTRACT_VERSION = self.BANCOR_ARBITRAGE_CONTRACT.caller.version()
+        self.ARB_REWARDS_PPM = self.BANCOR_ARBITRAGE_CONTRACT.caller.rewards()[0]
```
sklbancor, line 188: I suppose you've done this alright, but it looks from the old code that self.BANCOR_ARBITRAGE_CONTRACT.caller.rewards()[0] was a string yet for you it is an int?

fastlane_bot/events/utils.py
```diff
@@ -1998,7 +1998,7 @@ def handle_tokens_csv(mgr, prefix_path, read_only: bool = False):
     )
 
 
-def check_and_approve_tokens(tokens: List, cfg) -> bool:
+def check_and_approve_tokens(tokens: List, cfg: Config):
```
sklbancor, line 2001: return type? it used to return bool, no longer?

resources/FAQ.md
```diff
@@ -6,7 +6,7 @@
 
 ### **2. Using a Private Node or Infura with the Fastlane Bot**
 - **Question:** Can a private node or Infura be used with the Fastlane Bot?
-- **Answer:** Yes. The bot can use any RPC. This can be edited in the fastlane_bot/config/providers file - search for: self.RPC_URL. Note that some functions require an Alchemy API key, making it necessary unless the functions themselves are modified. These functions are in fastlane_bot/helpers/txhelpers, and include get_access_list, submit_private_transaction, and get_max_priority_fee_per_gas_alchemy.
+- **Answer:** Yes. The bot can use any RPC. This can be edited in the fastlane_bot/config/providers file - search for: self.RPC_URL. Note that some functions require an Alchemy API key, making it necessary unless the functions themselves are modified. These functions are in fastlane_bot/helpers/txhelpers, and include _create_access_list, _send_private_transaction, and _max_priority_fee_per_gas.
```
sklbancor, line 9: we should not refer to private functions in the documentation; by Python convention, functions starting with underscore should be considered private, meaning they are not part of the API and outsiders should not call them

resources/How_to_make_your_bot_competitive.md
```diff
@@ -13,9 +13,9 @@ Faster data means faster cycles for the arbitrage bot. The easiest way to achiev
 Another way to achieve this is by running an Ethereum node locally. The bot can be connected to a local Ethereum node fairly easily by changing the RPC_URL in the fastlane_bot/config/providers file. Note that some functions require an Alchemy API key, making Alchemy necessary unless the functions themselves are modified. 
 
 These functions are in fastlane_bot/helpers/txhelpers, and include: 
-* **get_access_list:** this function is optional - using it saves around 5000 gas on average per transaction.
-* **submit_private_transaction:** this function is not optional, however it's possible to submit transactions directly to Flashbots. See the Flashbots Documentation for more details: https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint
-* **get_max_priority_fee_per_gas_alchemy:** this function is not optional, but could be replaced with custom priority fee logic. 
+* **_create_access_list:** this function is optional - using it saves around 5000 gas on average per transaction.
```
sklbancor, line 16: see above comment regarding private functions

resources/FAQ.md
barakman: Initially I indeed just removed this part altogether, but then felt that it was nevertheless useful.
Neither one of these functions is used outside of that module, hence I have "privatized" them.
Removing the underscore would hide that fact, while keeping it makes this description flawed.
But this is a purely semantic issue, so resolving.

resources/How_to_make_your_bot_competitive.md
barakman: Same as above

fastlane_bot/config/provider.py
barakman: Yes, no point in converting it to a string when it is later used for arithmetic computation.
You'd also note that the name 'percentage' was wrong, since it is a normalized value (between 0 and 1).
The correct naming would be 'portion', but I went further ahead and preserved the original (on-chain) resolution.
So it is now a PPM integer value, and the division by 1 million takes place during the actual computations.

fastlane_bot/events/utils.py
barakman: Wasn't used before either, so I removed it completely.

sklbancor (dismissed):

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +54,150 @@ def validate_and_submit_transaction(
...
-        """Handles exceptions that occur during transaction building.
+        gas_cost_wei = int(gas_price * estimated_gas * self.cfg.EXPECTED_GAS_MODIFIER)
+        if self.cfg.network.GAS_ORACLE_ADDRESS:
+            gas_cost_wei += self.cfg.GAS_ORACLE_CONTRACT.caller.getL1Fee(raw_tx)
 
-        This method logs the exception and attempts to extract information from it. If the exception
-        indicates that the maximum fee per gas is less than the block base fee, it extracts and returns
-        the base fee. Otherwise, it logs a warning and returns None.
+        gas_cost_eth = Decimal(gas_cost_wei) / ETH_DECIMALS
+        gas_cost_usd = gas_cost_eth * expected_profit_usd / expected_profit_gastkn
```
sklbancor, line 115: that's ugly -- why don't you have as parameter a expected profit figure and an exchange rate, rather than having to expected profit figures that are then used to back out the exchange rate?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -34,81 +31,21 @@ class TxHelpers:
...
-
-        # Multiply by reward percentage, taken from the arb contract
-        adjusted_reward_eth = Decimal(Decimal(expected_profit_eth) * Decimal(self.ConfigObj.ARB_REWARD_PERCENTAGE))
-        adjusted_reward_usd = adjusted_reward_eth * expected_profit_usd / expected_profit_eth
-
-        return gas_cost_eth, gas_cost_usd, adjusted_reward_eth, adjusted_reward_usd
+        self.arb_contract = self.cfg.BANCOR_ARBITRAGE_CONTRACT
+        self.use_tenderly = self.cfg.NETWORK == self.cfg.NETWORK_TENDERLY
+        self.use_eip_1559 = self.cfg.NETWORK in ["ethereum", "coinbase_base"]
+        self.access_lists = self.cfg.NETWORK in ["ethereum"]
```
sklbancor, line 40: calling a `bool` variable `access_lists` is misleading; `had_access_lists` or `requires_access_lists` or something like that?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -34,81 +31,21 @@ class TxHelpers:
...
-        gas_cost_usd = gas_cost_eth * expected_profit_usd / expected_profit_eth
-
-        # Multiply by reward percentage, taken from the arb contract
-        adjusted_reward_eth = Decimal(Decimal(expected_profit_eth) * Decimal(self.ConfigObj.ARB_REWARD_PERCENTAGE))
-        adjusted_reward_usd = adjusted_reward_eth * expected_profit_usd / expected_profit_eth
-
-        return gas_cost_eth, gas_cost_usd, adjusted_reward_eth, adjusted_reward_usd
+        self.arb_contract = self.cfg.BANCOR_ARBITRAGE_CONTRACT
+        self.use_tenderly = self.cfg.NETWORK == self.cfg.NETWORK_TENDERLY
+        self.use_eip_1559 = self.cfg.NETWORK in ["ethereum", "coinbase_base"]
```
sklbancor, line 39: why are we using `"ethereum"` etc when `NETWORK_ETHEREUM` is defined in selectors.py

fastlane_bot/helpers/txhelpers.py
```diff
@@ -34,81 +31,21 @@ class TxHelpers:
...
+        self.access_lists = self.cfg.NETWORK in ["ethereum"]
+        self.arb_rewards = Decimal(self.cfg.ARB_REWARDS_PPM) / 1_000_000
+        self.eth = self.cfg.w3.eth
```
sklbancor, line 42: what is `self.eth`? admittedly the original sin may be in the web3 library naming convention and it is the "eth API"

https://web3py.readthedocs.io/en/stable/web3.eth.html

however -- especially for non-ETH chains this kinda does not make sense in web3, and it does not make sense here

fastlane_bot/helpers/txhelpers.py
```diff
@@ -34,81 +31,21 @@ class TxHelpers:
...
+
+        if self.use_tenderly:
+            self.wallet_address = self.cfg.BINANCE14_WALLET_ADDRESS
```
sklbancor, line 45: that's ugly; I don't think we can change it here because of our overall data structure, but changing the wallet address based on the network is ugly

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +54,150 @@ def validate_and_submit_transaction(
...
-        This method logs the exception and attempts to extract information from it. If the exception
-        indicates that the maximum fee per gas is less than the block base fee, it extracts and returns
-        the base fee. Otherwise, it logs a warning and returns None.
+        gas_cost_eth = Decimal(gas_cost_wei) / ETH_DECIMALS
+        gas_cost_usd = gas_cost_eth * expected_profit_usd / expected_profit_gastkn
 
-        Args:
-            self: The instance of the class.
-            exception (Exception): The exception raised during transaction building.
+        gas_gain_eth = self.arb_rewards * expected_profit_gastkn
```
sklbancor, line 117: I don't understand this; not saying it is wrong -- but what is this arb_rewards number? I understand it is set constant in the cfg object?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +54,150 @@ def validate_and_submit_transaction(
...
-        Returns:
-            Decimal: the expected layer one gas fee
-        """
-
-        l1_data_fee = asyncio.get_event_loop().run_until_complete(
-            asyncio.gather(self.ConfigObj.GAS_ORACLE_CONTRACT.caller.getL1Fee(raw_transaction)))[0]
-        # Dividing by 10 ** 18 in order to convert from wei resolution to native-token resolution
-        return Decimal(f"{l1_data_fee}e-18")
+            self.cfg.logger.info(f"Transaction {tx_hash} completed")
+        except TimeExhausted as _:
```
sklbancor, line 202: where is TimeExhausted defined?

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +54,150 @@ def validate_and_submit_transaction(
...
-            {
-                "tx": signed_tx.rawTransaction.hex(),
-                "maxBlockNumber": hex(block_number + 10),
-                "preferences": {"fast": True},
-            }
-        ]
+            self.cfg.logger.info(f"Failed building transaction {tx_details}; exception {e}")
+            return None, gas_price
 
+    def _send_private_transaction(self, raw_tx):
```
sklbancor, line 188: doc string

fastlane_bot/helpers/txhelpers.py
```diff
@@ -34,81 +31,21 @@ class TxHelpers:
...
+        self.arb_rewards = Decimal(self.cfg.ARB_REWARDS_PPM) / 1_000_000
+        self.eth = self.cfg.w3.eth
+
+        if self.use_tenderly:
+            self.wallet_address = self.cfg.BINANCE14_WALLET_ADDRESS
+        else:
+            self.wallet_address = str(self.eth.account.from_key(self.cfg.ETH_PRIVATE_KEY_BE_CAREFUL).address)
+            self.alchemy = Alchemy(api_key=self.cfg.WEB3_ALCHEMY_PROJECT_ID, network=Network.ETH_MAINNET, max_retries=3)
 
     def validate_and_submit_transaction(
```
sklbancor, line 50: doc string should have parameter list; consider converting all dosstrings to Google format

sklbancor (changes requested):
most changes are minor; if we do type hints private functions should also have them, and functions that return None should be marked `-> None`; also all methods should have docstrings, including the private ones

fastlane_bot/helpers/txhelpers.py
```diff
@@ -7,23 +7,20 @@
...
 from web3.exceptions import TimeExhausted
 
 from fastlane_bot.config import Config
 from fastlane_bot.data.abi import ERC20_ABI
-from fastlane_bot.utils import num_format, log_format, num_format_float, int_prefix
+from fastlane_bot.utils import num_format, log_format
 
-nest_asyncio.apply()
+MAX_UINT256 = 2 ** 256 - 1
+ETH_DECIMALS = 10 ** 18
```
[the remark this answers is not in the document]
barakman: I wanted to avoid the long name `ETH_RESOLUTION`, and I was hoping that no one would notice :)

Please suggest a name if you can think of something better than `ETH_RESOLUTION`.

fastlane_bot/helpers/txhelpers.py
barakman: I kept the previous implementation here.

I will check to see if the caller may as well pass already-processed input (BTW, I have done that for a bunch of other input parameters, sometimes even removing them altogether, but not for this one).

fastlane_bot/helpers/txhelpers.py
barakman: Kept the previous implementation as is, but good point, will fix.

fastlane_bot/helpers/txhelpers.py
barakman: It is an object which exposes many of the web3 functions being used in this module.
Using it allows to slightly reduce the code length, but if you consider this a bad practice then I fix it.

fastlane_bot/helpers/txhelpers.py
barakman: So I take it that you don't see all variable names being of the same length as a high-priority objective?

fastlane_bot/helpers/txhelpers.py
barakman: I kept the previous implementation here (or more accurately - improved it a lot, but left this address in use).

I agree with in being ugly, and moreover, something that the user can simply pass if they choose to.

But it would require verifying that our Tenderly users (mostly Nick, that is) are happy with it.

fastlane_bot/helpers/txhelpers.py
barakman: A normalized portion (a value between 0 and 1).

fastlane_bot/helpers/txhelpers.py
barakman: At the top of this file.

fastlane_bot/helpers/txhelpers.py
barakman: So I take it that you don't see all variable names being of the same length a high-priority objective?

fastlane_bot/helpers/txhelpers.py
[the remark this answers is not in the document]
sklbancor: ETHDECIMALSMULTIPLIER

fastlane_bot/helpers/txhelpers.py
sklbancor: I prefer readable variables, and especially not misleading ones.

fastlane_bot/helpers/txhelpers.py
barakman: Fixed

fastlane_bot/helpers/txhelpers.py
sklbancor: it is not bad practice it is just a bad name; what it really is is a provider, and it is not necessarily for ETH. not a killer but I'd prefer a more meaningful name

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +54,150 @@ def validate_and_submit_transaction(
...
         expected_profit_gastkn: Decimal,
         expected_profit_usd: Decimal,
-        verbose: bool = False,
-        safety_override: bool = False,
-        log_object: Dict[str, Any] = None,
-        flashloan_struct: List[Dict[str, int or str]] = None,
-    ) -> Optional[Dict[str, Any]]:
+        log_object: Dict[str, Any],
+        flashloan_struct: List[Dict]
+    ) -> str:
```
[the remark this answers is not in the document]
sklbancor: @barakman

fastlane_bot/helpers/txhelpers.py
sklbancor: ok

fastlane_bot/helpers/txhelpers.py
sklbancor: yeah. just flagging. don't really have a local fix for that.

fastlane_bot/helpers/txhelpers.py
sklbancor: hm. yeah, not clear. a comment may be helpful.

fastlane_bot/helpers/txhelpers.py
sklbancor: huh. my IDE said "undefined".

sklbancor (approved):

fastlane_bot/helpers/txhelpers.py
barakman: Yes, that wasn't a serious question; fixed.

fastlane_bot/helpers/txhelpers.py
barakman: Replaced with the original (`self.cfg.w3.eth)`.

fastlane_bot/helpers/txhelpers.py
barakman: Renamed to `arb_rewards_portion`.

fastlane_bot/helpers/txhelpers.py
barakman: @NIXBNT:

sklbancor (approved):

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +53,136 @@ def validate_and_submit_transaction(
...
-        if self.ConfigObj.SELF_FUND:
-            return self.arb_contract.functions.fundAndArb(
-                routes, src_address, src_amt
-            )
-        return self.arb_contract.functions.flashloanAndArbV2(
-            flashloan_struct, routes
-        )
+        raw_tx = self.cfg.w3.eth.account.sign_transaction(tx, self.cfg.ETH_PRIVATE_KEY_BE_CAREFUL).rawTransaction
 
-    def _handle_function_build_exception(self, exception: Exception) -> int or None:
```
Lesigh-3100, line 291: Removing `_handle_function_build_exception` is extremely sub-optimal. The exception returns a valid baseFee, letting the bot proceed with what could be a perfectly valid arb opportunity without rerunning the entire cycle. By forcing the cycle to restart we have a high likelihood of losing the opportunity due to the time it takes to rerun the cycle.

fastlane_bot/events/utils.py
```diff
@@ -1998,93 +1998,13 @@ def handle_tokens_csv(mgr, prefix_path, read_only: bool = False):
     )
 
 
-def self_funding_warning_sequence(cfg):
```
Lesigh-3100, line 2001: Removing the warning should be a product decision. @ashachaf WDYT?

fastlane_bot/events/utils.py
```diff
@@ -1998,93 +1998,13 @@ def handle_tokens_csv(mgr, prefix_path, read_only: bool = False):
...
-        return True
-
-    for _tkn in unapproved_tokens:
-        tx = tx_helpers.approve_token_for_arb_contract(token_address=_tkn)
-        if tx is not None:
-            continue
-        else:
-            assert (
-                False
-            ), f"Failed to approve token: {_tkn}. This can be fixed by approving manually, or restarting the bot to try again."
```
Lesigh-3100, line 2085: I disagree with removing this check entirely because if any of the approval transactions fail, it will result in the bot appearing to be running normally but being unable to succeed in arbitrage transactions. 

I have no problem with moving this assertion to the new `check_and_approve_tokens` though - for example just adding an assertion that the TX was successful.

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +53,136 @@ def validate_and_submit_transaction(
...
+            f"- Expected cost: {num_format(gas_cost_eth)} GAS token ({num_format(gas_cost_usd)} USD)\n"
+            f"- Expected gain: {num_format(gas_gain_eth)} GAS token ({num_format(gas_gain_usd)} USD)\n"
         )
-        if "max fee per gas less than block base fee" in str(exception):
-            message = str(exception)
-            return int_prefix(message.split("baseFee: ")[1])
 
+        if gas_gain_eth > gas_cost_eth:
+            self.cfg.logger.info("Attempting to execute profitable arb transaction")
+            try:
```
Lesigh-3100, line 115: `_send_private_transaction` can only be used on Ethereum.

All other chains + Tenderly require the normal transaction submission. 

Currently here only Tenderly uses the normal transaction submission & all other chains use the private method - which will fail except on Ethereum.

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +52,129 @@ def validate_and_submit_transaction(
...
-                "maxBlockNumber": hex(block_number + 10),
-                "preferences": {"fast": True},
-            }
-        ]
 
+    def _send_private_transaction(self, raw_tx):
         response = self.alchemy.core.provider.make_request(
             method="eth_sendPrivateTransaction",
-            params=params,
+            params=[{"tx": raw_tx, "maxBlockNumber": self.cfg.w3.eth.block_number + 10, "preferences": {"fast": True}}],
```
Lesigh-3100, line 168: tx & maxBlockNumber are submitted here as hex. See original function.
image

fastlane_bot/helpers/txhelpers.py
```diff
@@ -117,578 +52,129 @@ def validate_and_submit_transaction(
...
+            if allowance == 0:
+                function = token_contract.functions.approve(self.arb_contract.address, MAX_UINT256)
+                tx = self._build_transaction(function=function, value=0)
+                if tx is not None:
+                    raw_tx = self.cfg.w3.eth.account.sign_transaction(tx, self.cfg.ETH_PRIVATE_KEY_BE_CAREFUL).rawTransaction
+                    tx_hash = self.cfg.w3.eth.send_raw_transaction(raw_tx)
+                    self._wait_for_transaction_receipt(tx_hash)
+
+    def _build_transaction(self, function, value):
+        tx_details = {
```
Lesigh-3100, line 149: Do you know if all chains currently supported & planned support EIP 1559 transactions? I recommend retaining it unless you are confident it's no longer in use

resources/How_to_make_your_bot_competitive.md
```diff
@@ -4,18 +4,10 @@
 
 Typically only one bot-operator is able to close each arbitrage opportunity, making bot operation competitive in nature. The purpose of this document is to provide ideas on ways to improve your own bot. 
 
-### Gas Priority Fee Optimization
-In config/network.py there is a constant: DEFAULT_GAS_PRICE_OFFSET. This is a multiplier that increases the priority fee paid by your TX. The higher it is, the more likely your TX will get executed, but the more the TX costs. It’s set to increase by 9% by default, which is likely insufficient to be competitive. You could also design a custom function to modify the priority fee based on profit - this could likely make you competitive, but would require some work.
```
Lesigh-3100, line 8: This section should be updated - not removed - as priority fee optimization is still an important part of competitiveness.

resources/How_to_make_your_bot_competitive.md
```diff
@@ -4,18 +4,10 @@
...
-
 ### Data Throughput
 Faster data means faster cycles for the arbitrage bot. The easiest way to achieve this is by using a premium plan from a data provider such as Alchemy. 
 
-Another way to achieve this is by running an Ethereum node locally. The bot can be connected to a local Ethereum node fairly easily by changing the RPC_URL in the fastlane_bot/config/providers file. Note that some functions require an Alchemy API key, making Alchemy necessary unless the functions themselves are modified. 
-
-These functions are in fastlane_bot/helpers/txhelpers, and include: 
-* **get_access_list:** this function is optional - using it saves around 5000 gas on average per transaction.
-* **submit_private_transaction:** this function is not optional, however it's possible to submit transactions directly to Flashbots. See the Flashbots Documentation for more details: https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint
```
Lesigh-3100, line 17: **submit_private_transaction:** This should not be removed as it's an important for running the bot with a local node.

Lesigh-3100 (changes requested):

fastlane_bot/helpers/txhelpers.py
barakman: This error happens when the block's `baseFee` becomes higher than the transaction's `maxGasFee`, which theoretically yields a negative tip to the miner, hence the transaction is immediately reverted. Recall that the miner gets paid `maxGasFee - baseFee`, regardless of the specified `maxPriorityFee`, which indicates the maximum tip and not the actual tip. But even before this happens, the block's `baseFee` may become sufficiently high to bring the effective tip paid to the miner extremely small, thus leading the transaction to linger for a very long time, and most likely become invalid (arbitrage already fulfilled) by the time it gets executed. So the only difference between these two scenarios is a reverted transaction vs a stuck transaction. Both can sporadically happen, just like front-running can sporadically invalidate arbs submitted by this bot, for example. This error can be handled in a "natural manner", at the next iteration of the bot's infinite loop, hence there is no need to tailor-make a special "handle error and retry" method for this specific error.

fastlane_bot/events/utils.py
barakman: The `check_and_approve_tokens` throws an exception in case of a failure (and the exact exception, not an assertion that we we make up, as in the code that this comment refers to).

fastlane_bot/helpers/txhelpers.py
barakman: Please suggest a robust non-hackie way for distinguishing between these two cases.

fastlane_bot/helpers/txhelpers.py
barakman: It doesn't need to be.
A 10-base string, a 16-base string starting with "0x", and a native number are all fine.

fastlane_bot/helpers/txhelpers.py
barakman: All chains currently supported in the bot (Ethereum, Base, Fantom and Mantle) support EIP1559.
Unused code (legacy-transaction support, in this case) should be removed.

resources/How_to_make_your_bot_competitive.md
barakman: The increase-factor has been removed, because it doesn't have a meaningful functionality.
See bullet 1 in the description of this PR for more details.
Subsequently, this description becomes irrelevant.

resources/How_to_make_your_bot_competitive.md
barakman: This function is internal in the `TxHelpers` class.
It is used whenever it can be used, and a bot user doesn't need to know anything about it.

fastlane_bot/helpers/txhelpers.py
barakman: Removed.

fastlane_bot/helpers/txhelpers.py
barakman: OK, resolving;

resources/How_to_make_your_bot_competitive.md
Lesigh-3100: The description became irrelevant, but the topic is still extremely relevant. 

Can we leave it with something like:
Creating custom handling for the bot's max priority fee can increase your bot's competitiveness.

fastlane_bot/helpers/txhelpers.py
Lesigh-3100: Native number is not fine according to documentation.

resources/How_to_make_your_bot_competitive.md
Lesigh-3100: Unless the `_send_private_transaction` function is changed to use the Flashbots API directly (which would be great), this point is still very important for someone who does not use Alchemy.

fastlane_bot/helpers/txhelpers.py
barakman: Fixed, TX!

resources/How_to_make_your_bot_competitive.md
Lesigh-3100: I agree with your analysis regarding adding a more configurable gas strategy, however, it's not necessary for this PR. 

My main concern here is informing bot-operators that this is something they can customize to be more competitive. 

Even a simple sentence that informs them of this factor is sufficient for now from my perspective.

barakman: Updated; please review

barakman:
Closing and reposting as #529

PR #480, opened 2024-03-22
fixes `tkn0 or tkn1 not found: None, None`

mikewcasale:
Fixes a bug introduced during dev->rc PR review. A reviewer commented how unreadable the return type-hint was for function `get_token_and_fee` and that it would be better to encapsulate the return in a new class, which I agreed with. However, when making the change, the return object mis-named the token attributes. Later on in the processing when the original names were searched and not found, this was logged but no hard errors thrown because in fact there are legitimate reasons that this same behavior could happen. 

This PR corrects the token attribute names of the respective class.

### DIFF
--- a/fastlane_bot/events/async_event_update_utils.py
+++ b/fastlane_bot/events/async_event_update_utils.py
@@ -88,8 +88,8 @@ class TokenFeeResponse:
     """
     exchange_name: str
     address: str
-    tkn0: str
-    tkn1: str
+    tkn0_address: str
+    tkn1_address: str
     fee: float
     cid: str
     strategy_id: int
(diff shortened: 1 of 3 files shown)


barakman (approved):

zavelevsky (dismissed):
LGTM

fastlane_bot/events/async_event_update_utils.py
```diff
@@ -100,20 +100,7 @@ def df(self) -> pd.DataFrame:
...
-                    "tkn0_address": self.tkn0,
-                    "tkn1_address": self.tkn1,
-                    "fee": self.fee,
-                    "cid": self.cid,
-                    "strategy_id": self.strategy_id,
-                    "anchor": self.anchor,
-                }
-            ]
-        )
+        return pd.DataFrame([self.__dict__])
```
sklbancor, line 103: wouldn't it be better to use `dataclass.asdict` here? that is imo the canonical way to converting a dataclass into a dict (importantly, it ignores non-dataclass properties should they ever be added)

sklbancor (changes requested):
also in the spirit of our recent discussion on Telegram -- maybe we should introduce a test that ensures that this error does not creep up on us again?

fastlane_bot/events/async_event_update_utils.py
barakman: I had a similar thought about `__dict__` myself (more precisely, about the fact that it probably shouldn't be accessed directly). But I was unaware of the `asdict` option, which sounds like a right way to go about.

mikewcasale: I have changed to use asdict and also have added unit tests.

PR #483, opened 2024-03-24
Update static data mantle

Lesigh-3100:
Updated static data for Mantle. 

Only significant change should be fees in Mantle's static_pool_data.csv.


zavelevsky (approved):
LGTM

PR #485, opened 2024-03-24
bugfix for get_tokens_and_fee issues

mikewcasale:
Fixes a bug in the get_tokens_and_fee function which tries to access a variable that wasn't created (due to name changes) - (i.e. it breaks the bot).


zavelevsky (approved):
LGTM

PR #486, opened 2024-03-25
Fix Cleopatra Contract Interaction

barakman:
Partially handles issue #488


fastlane_bot/events/pools/solidly_v2.py
```diff
@@ -88,11 +88,13 @@ def update_from_contract(
...
 
-        try:
+        if hasattr(contract.caller, "factory"):
             factory_address = contract.caller.factory()
-        except Exception:
-            # Velocimeter does not expose factory function - call voter to get an address that is the same for all Velcoimeter pools
+        elif hasattr(contract.caller, "voter"):
             factory_address = contract.caller.voter()
+        else:
+            slot = contract.w3.eth.get_storage_at(contract.address, 11)
```
sklbancor, line 96: can you please add a comment here what this does? as a reader I can kind follow the factory / voter things but this is just black magic

fastlane_bot/events/pools/solidly_v2.py
```diff
@@ -120,11 +123,13 @@ async def async_update_from_contract(
...
         reserve_balance = await contract.caller.getReserves()
 
-        try:
+        if hasattr(contract.caller, "factory"):
             factory_address = await contract.caller.factory()
-        except Exception:
-            # Velocimeter does not expose factory function - call voter to get an address that is the same for all Velcoimeter pools
+        elif hasattr(contract.caller, "voter"):
             factory_address = await contract.caller.voter()
+        else:
```
sklbancor, line 130: ditto

sklbancor (dismissed):
seems simple enough so it should work; please do add the comments though

fastlane_bot/events/pools/solidly_v2.py
barakman: Fixed

fastlane_bot/events/pools/solidly_v2.py
barakman: Fixed

sklbancor (approved):

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -83,13 +83,14 @@ async def get_fee_4(address: str, contract: Contract, factory_contract: Contract
...
 
+
 EXCHANGE_INFO = {
     "velocimeter_v2": {"decimals": 4, "factory_abi": VELOCIMETER_V2_FACTORY_ABI, "pool_abi": VELOCIMETER_V2_POOL_ABI, "fee_function": get_fee_1},
     "equalizer_v2": {"decimals": 4, "factory_abi": SCALE_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_2},
     "aerodrome_v2": {"decimals": 4, "factory_abi": SOLIDLY_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_3},
     "velodrome_v2": {"decimals": 4, "factory_abi": SOLIDLY_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_3},
     "scale_v2": {"decimals": 18, "factory_abi": SCALE_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_2},
-    "cleopatra_v2": {"decimals": 4, "factory_abi": CLEOPATRA_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_4},
+    "cleopatra_v2": {"decimals": 18, "factory_abi": CLEOPATRA_V2_FACTORY_ABI, "pool_abi": CLEOPATRA_V2_POOL_ABI, "fee_function": get_fee_4},
```
Lesigh-3100, line 93: "decimals": 18 is wrong, the current value 4 is correct.

fastlane_bot/events/pools/solidly_v2.py
```diff
@@ -120,11 +126,16 @@ async def async_update_from_contract(
         """
         reserve_balance = await contract.caller.getReserves()
 
-        try:
+        if hasattr(contract.caller, "factory"):
```
Lesigh-3100, line 129: This looks good.

Can we make it so the "else" statement can fail & add the ability to continue even if factory_address is None/empty string?

Lesigh-3100 (changes requested):

fastlane_bot/events/exchanges/solidly_v2.py
barakman: https://explorer.mantle.xyz/address/0xba94D26761E038163F04F7B2c7a5D232492cfbfD?tab=read_proxy

Click on function `decimals` (number 6 on the list), and you'll see 18:
image

fastlane_bot/events/pools/solidly_v2.py
barakman: 1. The `else` statement will never end up setting `factory_address` to "None/empty string", as it reads the 256-bit value stored at slot 11, and then truncates it to a 160-bit address value. So for a non-cleopatra pool, it would just return some random address (whatever's stored at slot 11 of the given pool contract).

2. Even if it did return some "None/empty string", how exactly would you continue? The code further down the road needs the address of a valid factory contract, so it would just revert somewhere else along the way, as soon as it attempted to interact with that contract.

fastlane_bot/events/exchanges/solidly_v2.py
barakman: OK, my bad on the fee-resolution.
I was looking into the contract's decimals function, which of course has nothing to do with the fees.
That resolution appears to be hard-coded (either in the factory contract or in the pool contract or both).
So I'm gonna assume that you've carefully investigated that in order to determine that it is indeed 4 and not 18.

fastlane_bot/events/pools/solidly_v2.py
```diff
@@ -88,11 +88,16 @@ def update_from_contract(
...
+            # in solidly, the factory is available via function `factory`
             factory_address = contract.caller.factory()
-        except Exception:
-            # Velocimeter does not expose factory function - call voter to get an address that is the same for all Velcoimeter pools
+        elif hasattr(contract.caller, "voter"):
+            # in velocimeter, the factory is available via function `voter`
             factory_address = contract.caller.voter()
+        else:
```
sklbancor, line 97: shouldn't we check for cleopatra here rather than assuming it is and possibly continue with a completely wrong value?

barakman: Correct.
Problem is, there is no way to tell whether or not the given contract is a Cleopatra pool.
BTW, the same problem exists prior to this PR (attempting to call contract function 'voter' as a fallback).
And due to the issue presented in a Stratum pool contract, where the address of the corresponding factory contract is not available in storage at all (neither via a contract function, nor by directly reading the contract storage), it seems that an overall different solution is required in the long run.

fastlane_bot/events/pools/solidly_v2.py
barakman: Updated the `else` part such that if the `factory_address` is not the address of SOME contract (i.e., no bytecode deployed at that address), it is set to the zero-address.
Of course, there is not a huge gain in this, other than making potential problems easier to investigate.
First off, because we might end up with `factory_address` pointing to some non-factory contract, whose address happens to be stored in slot 11 of the pool contract.
Second, because a zero address will nevertheless throw an exception as soon as we attempt to interact with a factory contract "deployed" (i.e., obviously not deployed) at that address.
So again - we should probably conceive a different solution altogether for this entire feature of obtaining factory addresses (e.g., store them hard-coded somewhere).

fastlane_bot/events/pools/solidly_v2.py
barakman: Updated the `else` part such that if the `factory_address` is not the address of SOME contract (i.e., no bytecode deployed at that address), it is set to the zero-address.
Of course, there is not a huge gain in this, other than making potential problems easier to investigate.
First off, because we might end up with `factory_address` pointing to some non-factory contract, whose address happens to be stored in slot 11 of the pool contract.
Second, because a zero address will nevertheless throw an exception as soon as we attempt to interact with a factory contract "deployed" (i.e., obviously not deployed) at that address.
So again - we should probably conceive a different solution altogether for this entire feature of obtaining factory addresses (e.g., store them hard-coded somewhere).

barakman:
Have decided on removing the attempt to fetch factory addresses dynamically (during runtime).

PR #491, opened 2024-03-26
support fusionx v3 on mantle

NIXBNT:


fastlane_bot/events/exchanges/uniswap_v3.py
```diff
@@ -30,7 +30,12 @@ def add_pool(self, pool: Pool):
         self.pools[pool.state["address"]] = pool
 
     def get_abi(self):
-        return UNISWAP_V3_POOL_ABI if self.exchange_name not in [PANCAKESWAP_V3_NAME, AGNI_V3_NAME] else PANCAKESWAP_V3_POOL_ABI
+        if self.exchange_name == FUSIONX_V3_NAME:
+            return FUSIONX_V3_POOL_ABI
+        elif self.exchange_name not in [PANCAKESWAP_V3_NAME, AGNI_V3_NAME]:
+            return UNISWAP_V3_POOL_ABI
+        else:
+            return PANCAKESWAP_V3_POOL_ABI
```
sklbancor, line 38: not specific to those changes here but this is a case where you should check for all possible conditions and have an else at the end that raises and "unknown exchange"

otherwise errors can percolate much further and may be harder to detect

sklbancor (approved):

sklbancor:
I based the branch on develop and it picked up a lot of weird stuff **DO NOT MOVE OUT OF DRAFT BEFORE DEALING WITH THIS**; talk to me after the RC is out

barakman:
> I based the branch on develop and it picked up a lot of weird stuff **DO NOT MOVE OUT OF DRAFT BEFORE DEALING WITH THIS**; talk to me after the RC is out

You might want to use that branch at the point in time when the RC branch was created, since many more changes have been pushed into it since (starting from around 22nd of March).

sklbancor:
> > I based the branch on develop and it picked up a lot of weird stuff **DO NOT MOVE OUT OF DRAFT BEFORE DEALING WITH THIS**; talk to me after the RC is out
> 
> 
> 
> You might want to use that branch at the point in time when the RC branch was created, since many more changes have been pushed into it since (starting from around 22nd of March).

Looking at the diffs I am pretty sure this PR is good.

sklbancor (approved):

Lesigh-3100 (approved):

barakman (approved):

fastlane_bot/events/exchanges/uniswap_v3.py
```diff
@@ -30,11 +30,11 @@ def add_pool(self, pool: Pool):
         self.pools[pool.state["address"]] = pool
 
     def get_abi(self):
-        return UNISWAP_V3_POOL_ABI if self.exchange_name not in [PANCAKESWAP_V3_NAME, AGNI_V3_NAME] else PANCAKESWAP_V3_POOL_ABI
+        return UNISWAP_V3_POOL_ABI if self.exchange_name not in [PANCAKESWAP_V3_NAME, AGNI_V3_NAME, FUSIONX_V3_NAME] else PANCAKESWAP_V3_POOL_ABI
 
     @property
     def get_factory_abi(self):
-        return UNISWAP_V3_FACTORY_ABI if self.exchange_name not in [PANCAKESWAP_V3_NAME, AGNI_V3_NAME] else PANCAKESWAP_V3_FACTORY_ABI
```
mikewcasale, line 37: technically, any `get_x()` method should not use an '@property' decorator, we should be following the example set by the base class for any abstract methods. However, it appears that there is only one usage of this method in the entire codebase, and that it's expecting '@property' behavior, so for the purposes of this PR I will approve this, but a new issue should be opened to correct this.

PR #492, opened 2024-03-26
Fix args kwargs usage in build transaction generic

Lesigh-3100:
This fixes a critical error in the `build_transaction_generic` function. The `**kwargs` were mistakenly passed into the contract function instead of the `build_tx` function in cases in which the initial gas base fee was wrong, causing it to fail.

I have verified that transactions are able to be constructed after this change.

### DIFF
--- a/fastlane_bot/helpers/txhelpers.py
+++ b/fastlane_bot/helpers/txhelpers.py
@@ -767,7 +767,7 @@ class TxHelpers:
                 return None
             try:
                 kwargs['gas_price'] = new_base_fee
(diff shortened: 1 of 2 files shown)


zavelevsky (approved):

fastlane_bot/helpers/txhelpers.py
sklbancor, line 1: Once the RC is out of the way the `*args, **kwargs` should be converted to`argsf1, argsf2` as discussed on Telegram

PR #495, opened 2024-03-27
Removed Factory address call for Solidly V2 pools

Lesigh-3100:
Removed call to fetch Factory address for Solidly V2 pools as the value was unused.


zavelevsky (approved):
LGTM

barakman (approved):

PR #505, opened 2024-03-28
Clean ABIs following the removal of the factory address retrieval

barakman:
1. Remove functions `factory` and `voter` from pool contract ABIs
2. Remove the Velocimeter-V2 ABI and use the Solidly-V2 ABI instead


fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -84,13 +84,13 @@ async def get_fee_4(address: str, contract: Contract, factory_contract: Contract
...
 
 EXCHANGE_INFO = {
-    "velocimeter_v2": {"decimals": 4, "factory_abi": VELOCIMETER_V2_FACTORY_ABI, "pool_abi": VELOCIMETER_V2_POOL_ABI, "fee_function": get_fee_1},
-    "equalizer_v2": {"decimals": 4, "factory_abi": SCALE_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_2},
-    "aerodrome_v2": {"decimals": 4, "factory_abi": SOLIDLY_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_3},
-    "velodrome_v2": {"decimals": 4, "factory_abi": SOLIDLY_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_3},
-    "scale_v2": {"decimals": 18, "factory_abi": SCALE_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_2},
-    "cleopatra_v2": {"decimals": 4, "factory_abi": CLEOPATRA_V2_FACTORY_ABI, "pool_abi": SOLIDLY_V2_POOL_ABI, "fee_function": get_fee_4},
-    "stratum_v2": {"decimals": 4, "factory_abi": VELOCIMETER_V2_FACTORY_ABI, "pool_abi": VELOCIMETER_V2_POOL_ABI, "fee_function": get_fee_1},
+    "velocimeter_v2": {"decimals": 4, "factory_abi": VELOCIMETER_V2_FACTORY_ABI, "fee_function": get_fee_1},
```
sklbancor, line 87: just because they are all the same currently will this always be the case? if not it sucks to undo those changes even with multi cursor editing

sklbancor (approved):

fastlane_bot/events/exchanges/solidly_v2.py
barakman: Same issue with all other solidly-like exchanges (equalizer, aerodrome, velodrome, scale, cleopatra, stratum, etc).
The only reason that the velocimeter ABI was added here in the first place was this specific difference (function 'voter' vs function 'factory'), which is no longer an issue.

If any such issue ever arises in the future (i.e., as a result of requiring some specific action which is different across different exchanges), then we will add it.

The general approach is to keep only the necessary code, otherwise maintenance continues to grow difficult.

barakman:
@sklbancor Please close when done inspecting :)

PR #511, opened 2024-03-29
Clean the `async_event_update_utils` module

barakman:
The main purpose in this PR is to get rid of the part:
```
    global cfg
    cfg = mgr.cfg
```
In function `async_update_pools_from_contracts`, which is therefore expected to be executed before any other function which relies on `cfg` being a global variable in this module.

This is done by making sure that every such function takes `mgr` as input.

Additional fixes included in this PR:
- Removing unused functions
- Removing unused input arguments
- Passing `mgr` as the first input argument in every function
- Prepending an underscore to the name of every internal function

---

Not handling any of the following semantic issues, in order to keep the primary changes in this PR easy to review:
- Type hints
- Doc strings


fastlane_bot/events/async_event_update_utils.py
```diff
@@ -155,19 +147,9 @@ async def main_get_tokens_and_fee(c: List[Dict[str, Any]]) -> pd.DataFrame:
...
-    return f"{t0_symbol}-{tkn0_address[-key_digits:]}/{t1_symbol}-{tkn1_address[-key_digits:]}"
-
-
-def get_pool_info(
-        pool: pd.Series,
+def _get_pool_info(
         mgr: Any,
+        pool: pd.Series,
         current_block: int,
         tkn0: Dict[str, Any],
```
sklbancor, line 154: why is tkn a dict?

fastlane_bot/events/async_event_update_utils.py
```diff
@@ -207,65 +189,9 @@ def get_pool_info(
     return pool_info
 
 
-def sanitize_token_symbol(token_symbol: str, token_address: str, read_only: bool) -> str:
```
sklbancor, line 210: why are those functions no longer needed?

fastlane_bot/events/async_event_update_utils.py
```diff
@@ -207,65 +189,9 @@ def get_pool_info(
...
-    try:
-        if not read_only:
-            token_pd = pd.DataFrame([{"symbol": token_symbol}], columns=["symbol"])
-            token_pd.to_csv(sanitization_path)
-        return token_symbol
-    except Exception:
-        return token_address
-
-
-def add_token_info(
```
sklbancor, line 231: ditto

fastlane_bot/events/async_event_update_utils.py
```diff
@@ -207,65 +189,9 @@ def get_pool_info(
...
-    pool_info["tkn1_symbol"] = tkn1_symbol
-    pool_info["tkn1_decimals"] = tkn1["decimals"]
-    pool_info["pair_name"] = tkn0["address"] + "/" + tkn1["address"]
-
-    if len(pool_info["pair_name"].split("/")) != 2:
-        raise Exception(f"pair_name is not valid for {pool_info}")
-    return pool_info
-
-
-def add_missing_keys(
```
sklbancor, line 253: ditto

sklbancor (approved):

sklbancor:
> The main purpose in this PR is to get rid of the part:
> 
> ```
> 
>     global cfg
> 
>     cfg = mgr.cfg
> 
> ```
> 
> In function `async_update_pools_from_contracts`, which is therefore expected to be executed before any other function which relies on `cfg` being a global variable in this module.
> 
> 
> 
> This is done by making sure that every such function takes `mgr` as input.
> 
> 
> 
> Additional fixes included in this PR:
> 
> - Removing unused functions
> 
> - Removing unused input arguments
> 
> - Passing `mgr` as the first input argument in every function
> 
> - Prepending an underscore to the name of every internal function
> 
> 
> 
> ---
> 
> 
> 
> Not handling any of the following semantic issues, in order to keep the primary changes in this PR easy to review:
> 
> - Type hints
> 
> - Doc strings

scope makes sense except I assume you ensure that when a function signature changes (eg, add manager) and a docstring exists this docstring is updated?

fastlane_bot/events/async_event_update_utils.py
barakman: I have no idea, but I've searched through the code, and they do not appear to be called.
I will appreciate your validation of this, of course.

fastlane_bot/events/async_event_update_utils.py
barakman: This is a question for another PR.
It probably appears in other places as well.
The only reason that you're seeing a change in this function, is because I moved `mgr` to be the first input argument, in order to keep a consistent coding-convention throughout this module (in fact, it might actually imply that all functions here should actually be implemented as part of the `Manager` class rather than as a set of global functions).

Issue #314, opened 2024-01-18
Fix the "fee,fee_float" scheme


barakman:
The current implementation of the "fee,fee_float" scheme is extremely inconsistent across different files and modules.
It also contains a lot of redundant information, as these two entities essentially represent the same value, and that too - reflects yet another inconsistency - in most cases,`fee` stands for the raw (scaled-up) fee in the contract, while `fee_float` stands for the normalized value of it (a value between 0 and 1), but in some rare cases, they actually hold the exact same value.

Very generally, there should be a static data file containing the fee-resolution of each pool (or even each DEX), for example:
- Carbon, 1e6
- Uniswap, 1e12
- Balancer, 1e18

But this issue is probably better off split into two phases, where that static data file will be added only during the 2nd phase:
- First phase: replace all occurrences of `fee_float` with `fee_resolution`
- Second phase: move the `fee_resolution` to be stored only in the static data file, instead of being replicated everywhere

The first phase should include setting the type of both `fee` and `fee_resolution` to `int`.
At present, the type of `fee` varies between `int`, `float` and `str` across different files.

The calculation of `fee_float` should really take place only at the very low level, where the real (normalized) value is required.
Preferably via `Decimal(fee) / fee_resolution`, but for starters, `float(fee) / fee_resolution` would also do fine.

This change is horizontal across core python modules, jupyter tests and static data files (all files named `static_pool_data.csv`).

PR #518, opened 2024-04-04
Add Nile V3 specific fee support

Lesigh-3100:
Nile V3 is a Uni V3 fork that has an override for the fee actually charged by the pool, however, the tick spacing correlated to the original fee remains the same.

This means that the bot must:

- Add handling to fetch the fee override & use it directly (divided by resolution 1e6) for the `fee_float` field in the TerraFormer & Exchange class.
- Use the `fee` field for everything related to Uniswap V3 ticks.
- Use the `fee` field for the customInt input to the Arb contract. 
- Use the `fee_float` field for exact calculations performed in `RouteHandler.py`.


run_blockchain_terraformer.py
```diff
@@ -466,6 +468,10 @@ def organize_pool_details_uni_v3(
         return None
     pair = pair[:-1]
     fee = pool_data["args"]["fee"]
+    fee_float = float(fee) / 1000000
+    if exchange == NILE_V3_NAME:
+        pool = web3.eth.contract(address=web3.to_checksum_address(pool_address), abi=NILE_V3_POOL_ABI)
+        fee_float = pool.caller.currentFee() / 1000000
```
barakman, line 474: Why function `currentFee` and not function `fee`?
For UniV3 pools, we are using function `fee`.
Might be worth verifying exactly which function is required here.
You can do that by looking into the source code of this pool contract, checking which fee is used in the swap function, and then following it to see which of these two functions returns that specific fee.

fastlane_bot/helpers/tradeinstruction.py
```diff
@@ -206,7 +206,7 @@ def get_custom_int(self) -> int:
         pool = self.pool
         custom_int = 0
         if self.exchange_name in self.ConfigObj.UNI_V3_FORKS:
-            custom_int = int(Decimal(pool.fee_float) * Decimal("1000000"))
+            custom_int = int(pool.fee)
```
barakman, line 209: How is this change related to this PR?

fastlane_bot/helpers/poolandtokens.py
```diff
@@ -520,10 +520,10 @@ def _univ3_to_cpc(self) -> List[Any]:
             "tick": self.tick,
             "liquidity": self.liquidity,
         }
-        feeconst = self.FEE_LOOKUP.get(float(self.fee_float))
+        feeconst = self.FEE_LOOKUP.get(int(self.fee)/1e6)
```
barakman, line 523: How is this change related to this PR?

fastlane_bot/helpers/poolandtokens.py
```diff
@@ -520,10 +520,10 @@ def _univ3_to_cpc(self) -> List[Any]:
...
         if feeconst is None:
             raise ValueError(
-                f"Illegal fee for Uniswap v3 pool: {self.fee_float} [{self.FEE_LOOKUP}]]"
+                f"Illegal fee for Uniswap v3 pool: {int(self.fee)} [{self.FEE_LOOKUP}]]"
```
barakman, line 526: How is this change related to this PR?

fastlane_bot/events/exchanges/uniswap_v3.py
```diff
@@ -42,6 +47,9 @@ def get_events(self, contract: Contract) -> List[Type[Contract]]:
     async def get_fee(self, address: str, contract: Contract) -> Tuple[str, float]:
         fee = await contract.functions.fee().call()
         fee_float = float(fee) / 1e6
+        if self.exchange_name == NILE_V3_NAME:
+            fee_float = await contract.functions.currentFee().call()
```
barakman, line 51: Why function `currentFee` and not function `fee`?
For UniV3 pools, we are using function `fee`.
Might be worth verifying exactly which function is required here.
You can do that by looking into the source code of this pool contract, checking which fee is used in the swap function, and then following it to see which of these two functions returns that specific fee.

run_blockchain_terraformer.py
barakman: And why you using `1000000` in some cases and `1e6` in other cases?

barakman (changes requested):

fastlane_bot/events/exchanges/uniswap_v3.py
barakman: Also, there is something really wrong here, as in one case you are dividing by the resolution, and in one case you don't.
So `fee_float` is sometimes returned after being divided by `16e`, and sometimes being returned as read from the contract.
How could that possibly work correctly further down the road, when the caller of this function receives this returned value and needs to processes it?
Also note that you seem to have made a similar change in file `run_blockchain_terraformer.py`, where you actually do divide by `1000000` in both cases.
So even without taking the above into consideration, it is pretty clear that at least one of these two changes is incorrect.

barakman (changes requested):

fastlane_bot/events/exchanges/uniswap_v3.py
barakman: In addition to that, it makes much more sense to implement a separate file called `nile_v3.py`.
This type of logic becomes increasingly confusing:
```
if self.exchange_name in [PANCAKESWAP_V3_NAME, AGNI_V3_NAME, FUSIONX_V3_NAME, ECHODEX_V3_NAME, SECTA_V3_NAME]:
    return PANCAKESWAP_V3_POOL_ABI
elif self.exchange_name == NILE_V3_NAME:
    return NILE_V3_POOL_ABI
return UNISWAP_V3_POOL_ABI
```
The guiding thumb-rule should be that once a simple branching (if/else) becomes too obfuscated, it is time to consider an alternative approach.
Note that an additional logic was also applied in function `get_fee` further down in this file, and on top of making it more cumbersome, this one actually introduces a bug:
```
fee_float = float(fee) / 1e6
if self.exchange_name == NILE_V3_NAME:
    fee_float = await contract.functions.currentFee().call()
```
As the value of `fee_float` is scaled down by `1e6` in some scenarios, and not scaled down at all in other scenarios.
So unless you apply the same logic (`if exchange_name == NILE_V3_NAME`) in every function which calls this function, an incorrect calculation most likely takes place.

run_blockchain_terraformer.py
barakman: Also, creating a `web3.eth.contract` instance clearly doesn't belong in here (as evident by how it is done elsewhere in the code. You can't just "throw stuff in" without taking into consideration the broader SW design and structure.

barakman: Also, it doesn't make sense that we need to make an RPC at this point, as the fee has already been obtained previously.
So during that previous point - either it was already obtained correctly using this contract function, or it there was an attempt to use the other function, which would have reverted for this contract.

fastlane_bot/events/exchanges/uniswap_v3.py
barakman: After a short research, the best approach here would be a paradigm similar to the one in solidly_v2.py.

run_blockchain_terraformer.py
Lesigh-3100: 1. The 1000000 vs 1e6 is just poor consistency - I will update this to use 1e6.
2. Nile V3 uses a fee override function - currentFee. This results in an awkward situation in the Terraformer, as in all other cases the bot obtains the fee from the pool creation event. Due to the override, the best way to obtain the pool's current fee is by directly calling the contract. 
3. Regarding the overall SW design, do you have a suggestion for a more appropriate place for this, given that it's only relevant for Nile V3? I'm happy to move it if it makes more sense elsewhere.

fastlane_bot/helpers/tradeinstruction.py
Lesigh-3100: Nile V3 includes a fee override.

This PR stores the fee override for the pool in the `fee_float` field, and the original fee (which is necessary for Uniswap V3 details such as tick-spacing etc) in the `fee` field.

The Fast Lane arb contract must know the `fee` of the Uni V3 pool being traded on, so using the `fee_float` to calculate the `fee` is no longer valid in instances of override. 

Consequently the `fee` field is now used as it does not change.

barakman: I did not understand this description.
But Nile V3 should not be handled differently than any other DEX, other than in the dedicated sections (for example, in the uniswap_v3 module, where all of these DEXs are handled).

run_blockchain_terraformer.py
barakman: Please see the PR spawned off of this one.
In order to handle this in the uniswap_v3 module, for example, I used a paradigm similar to the one implemented in the solidly_v2 module.

The part:
```
        pool = web3.eth.contract(address=web3.to_checksum_address(pool_address), abi=NILE_V3_POOL_ABI)
        fee_float = pool.caller.currentFee() / 1000000
```
Doesn't look like it belongs here.
Even without being acquainted with this module, a simple observation reveals that no other DEX requires an RPC for fetching the fee at this point, so why should the Nile V3 DEX require one?
Subsequently, placing it here breaks the structure of the code and the over design of this module.
Wherever the fee for all the other DEXs is fetched, that's where the fee for Nile V3 should be fetched.

fastlane_bot/helpers/tradeinstruction.py
Lesigh-3100: Correct, and to handle it the same, we must now use the `fee` field - which remains constant, instead of the `fee_float` which has an override for Nile V3.

fastlane_bot/events/exchanges/uniswap_v3.py
```diff
@@ -42,6 +47,9 @@ def get_events(self, contract: Contract) -> List[Type[Contract]]:
     async def get_fee(self, address: str, contract: Contract) -> Tuple[str, float]:
         fee = await contract.functions.fee().call()
         fee_float = float(fee) / 1e6
+        if self.exchange_name == NILE_V3_NAME:
+            fee_float = await contract.functions.currentFee().call() / 1e6
```
barakman, line 51: Potential risk of a runtime issue described in issue #314.

PR #528, opened 2024-04-09
Exclude known tax tokens

barakman:
Using a single list for all networks; possibly need to use one list per network.


fastlane_bot/bot.py
```diff
@@ -916,6 +916,19 @@ def _handle_trade_instructions(
...
+        ]
+
+        if not indices:
+            self.ConfigObj.logger.info(f"[bot._handle_trade_instructions] only tax tokens detected, discarding.")
+            return None
+
+        best_trade_instructions_dic = [x for n, x in enumerate(best_trade_instructions_dic) if n in indices]
+        best_trade_instructions_df = [x for n, x in enumerate(best_trade_instructions_df) if n in indices]
+        best_trade_instructions = [x for n, x in enumerate(best_trade_instructions) if n in indices]
+
```
barakman, line 931: TODO:
What if `best_profit` and `best_src_token` refer to an instructions which was just removed?

fastlane_bot/config/network.py
```diff
@@ -193,6 +193,12 @@ class ConfigNetwork(ConfigBase):
     BNT_ADDRESS = "0x1F573D6Fb3F13d689FF844B4cE37794d79a7FF1C"
     USDT_ADDRESS = "0xdAC17F958D2ee523a2206206994597C13D831ec7"
     WETH_ADDRESS = WETH9_ADDRESS = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+
+    TAX_TOKENS = [
+        "0x5a3e6A77ba2f983eC0d371ea3B475F8Bc0811AD5", # 0x0
+        "0x72e4f9F808C49A2a61dE9C5896298920Dc4EEEa9", # BITCOIN
+    ]
+
```
barakman, line 201: TODO: might need to define this per network.

fastlane_bot/bot.py
```diff
@@ -916,6 +916,19 @@ def _handle_trade_instructions(
             best_trade_instructions,
         ) = r
 
+        indices = [
```
zavelevsky, line 919: my main concern is that it's a late phase and I would expect us to be able to not include these tokens/pools in the data we pass to the optimizer

Lesigh-3100:
A better place to do this would be in the `handle_subsequent_iterations` function in `utils.py`. For example, if a blacklist were passed in as an argument to the `handle_subsequent_iterations ` function, it could be used similarly to: 
```
if target_tokens:
    bot.db.filter_target_tokens(target_tokens)
```

You could add a function similar to this, for example `filter_blacklisted_tokens`:

```
    def filter_blacklisted_tokens(self, blacklist: List[str]):
        """
        Filter the pools to remove pools that include blacklisted tokens.

        Parameters
        ----------
        blacklist: List[str]
            This list of tokens will be used to remove any pools that contain them. 
        """
        self.state = [
            pool
            for pool in self.state
            if pool["tkn0_address"] not in blacklistand pool["tkn1_address"] not in blacklist
        ]
```

This will result in removing any pools that contain the blacklisted tokens before being sent to the Optimizer.

fastlane_bot/bot.py
barakman: From what I've observed so far (as well as what I know), most of the execution time is spent on RPCs and not on purely internal computations (even the complicated ones taking place inside the optimizer).

Of course, this is not to say that we shouldn't strive to place this filter as early as possible within the process.

And regardless, the current solution is by itself most likely flawed, because as explained in[this comment of mine](https://github.com/bancorprotocol/fastlane-bot/pull/528/files#r1557935564) above, it might result with `best_profit` and `best_src_token` being "stale" values, referring to a token which has just been filtered out of those 3 trade-instruction lists.

So I'm now thinking of doing it function `get_curves`, immediately after the line:
```
tokens = self.db.get_tokens()
```

barakman:
See [this comment of mine](https://github.com/bancorprotocol/fastlane-bot/pull/528#discussion_r1560655139).

What about filtering out these tax tokens in function `get_curves`, immediately after the line:
```
tokens = self.db.get_tokens()
```
?

barakman:
> A better place to do this would be in the `handle_subsequent_iterations` function in `utils.py`. For example, if a blacklist were passed in as an argument to the `handle_subsequent_iterations ` function, it could be used similarly to:
> 
> ```
> if target_tokens:
>     bot.db.filter_target_tokens(target_tokens)
> ```
> 
> You could add a function similar to this, for example `filter_blacklisted_tokens`:
> 
> ```
>     def filter_blacklisted_tokens(self, blacklist: List[str]):
>         """
>         Filter the pools to remove pools that include blacklisted tokens.
> 
>         Parameters
>         ----------
>         blacklist: List[str]
>             This list of tokens will be used to remove any pools that contain them. 
>         """
>         self.state = [
>             pool
>             for pool in self.state
>             if pool["tkn0_address"] not in blacklistand pool["tkn1_address"] not in blacklist
>         ]
> ```
> 
> This will result in removing any pools that contain the blacklisted tokens before being sent to the Optimizer.

See [this comment of mine](https://github.com/bancorprotocol/fastlane-bot/pull/528#discussion_r1560655139).

What about filtering out these tax tokens in function `get_curves`, immediately after the line:
```
tokens = self.db.get_tokens()
```
?

Lesigh-3100:
> > A better place to do this would be in the `handle_subsequent_iterations` function in `utils.py`. For example, if a blacklist were passed in as an argument to the `handle_subsequent_iterations ` function, it could be used similarly to:
> > ```
> > if target_tokens:
> >     bot.db.filter_target_tokens(target_tokens)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You could add a function similar to this, for example `filter_blacklisted_tokens`:
> > ```
> >     def filter_blacklisted_tokens(self, blacklist: List[str]):
> >         """
> >         Filter the pools to remove pools that include blacklisted tokens.
> > 
> >         Parameters
> >         ----------
> >         blacklist: List[str]
> >             This list of tokens will be used to remove any pools that contain them. 
> >         """
> >         self.state = [
> >             pool
> >             for pool in self.state
> >             if pool["tkn0_address"] not in blacklistand pool["tkn1_address"] not in blacklist
> >         ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will result in removing any pools that contain the blacklisted tokens before being sent to the Optimizer.
> 
> See [this comment of mine](https://github.com/bancorprotocol/fastlane-bot/pull/528#discussion_r1560655139).
> 
> What about filtering out these tax tokens in function `get_curves`, immediately after the line:
> 
> ```
> tokens = self.db.get_tokens()
> ```
> 
> ?

The place you indicated is slightly less ideal because it's farther along in the process & results in additional computation for pool data that is unusable.

barakman:
> > > A better place to do this would be in the `handle_subsequent_iterations` function in `utils.py`. For example, if a blacklist were passed in as an argument to the `handle_subsequent_iterations ` function, it could be used similarly to:
> > > ```
> > > if target_tokens:
> > >     bot.db.filter_target_tokens(target_tokens)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > You could add a function similar to this, for example `filter_blacklisted_tokens`:
> > > ```
> > >     def filter_blacklisted_tokens(self, blacklist: List[str]):
> > >         """
> > >         Filter the pools to remove pools that include blacklisted tokens.
> > > 
> > >         Parameters
> > >         ----------
> > >         blacklist: List[str]
> > >             This list of tokens will be used to remove any pools that contain them. 
> > >         """
> > >         self.state = [
> > >             pool
> > >             for pool in self.state
> > >             if pool["tkn0_address"] not in blacklistand pool["tkn1_address"] not in blacklist
> > >         ]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This will result in removing any pools that contain the blacklisted tokens before being sent to the Optimizer.
> > 
> > 
> > See [this comment of mine](https://github.com/bancorprotocol/fastlane-bot/pull/528#discussion_r1560655139).
> > What about filtering out these tax tokens in function `get_curves`, immediately after the line:
> > ```
> > tokens = self.db.get_tokens()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> The place you indicated is slightly less ideal because it's farther along in the process & results in additional computation for pool data that is unusable.

OK, but `handle_subsequent_iterations` is executed even before the main loop.
Isn't there a chance that more tokens will be added after that, and a tax token will eventually find its way into the trade instructions?

Also, what is `self.state`?
I couldn't find it neither in function `handle_subsequent_iterations`, nor in the function which calls it.

Lesigh-3100:
> > > > A better place to do this would be in the `handle_subsequent_iterations` function in `utils.py`. For example, if a blacklist were passed in as an argument to the `handle_subsequent_iterations ` function, it could be used similarly to:
> > > > ```
> > > > if target_tokens:
> > > >     bot.db.filter_target_tokens(target_tokens)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > You could add a function similar to this, for example `filter_blacklisted_tokens`:
> > > > ```
> > > >     def filter_blacklisted_tokens(self, blacklist: List[str]):
> > > >         """
> > > >         Filter the pools to remove pools that include blacklisted tokens.
> > > > 
> > > >         Parameters
> > > >         ----------
> > > >         blacklist: List[str]
> > > >             This list of tokens will be used to remove any pools that contain them. 
> > > >         """
> > > >         self.state = [
> > > >             pool
> > > >             for pool in self.state
> > > >             if pool["tkn0_address"] not in blacklistand pool["tkn1_address"] not in blacklist
> > > >         ]
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > This will result in removing any pools that contain the blacklisted tokens before being sent to the Optimizer.
> > > 
> > > 
> > > See [this comment of mine](https://github.com/bancorprotocol/fastlane-bot/pull/528#discussion_r1560655139).
> > > What about filtering out these tax tokens in function `get_curves`, immediately after the line:
> > > ```
> > > tokens = self.db.get_tokens()
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ?
> > 
> > 
> > The place you indicated is slightly less ideal because it's farther along in the process & results in additional computation for pool data that is unusable.
> 
> OK, but `handle_subsequent_iterations` is executed even before the main loop. Isn't there a chance that more tokens will be added after that, and a tax token will eventually find its way into the trade instructions?
> 
> Also, this function seems to expect a possible scenario where no target tokens are provided as input (`target_tokens = None`), which implies that they are obtained at a later point in the process, in which case, a tax token can still "slip into the system" at that point... no?

FYI `target_tokens` are a user-input configuration option used to limit search scope by targeting arbitrage for specific tokens - they don't change during runtime. If no target tokens are specified - the bot searches all available pools.

`handle_subsequent_iterations` is called after data collection is complete - no new pools or tokens are added to the bot's data at this point. By removing pools with tax tokens here, we are removing them from all future computation in the bot's cycle. I think this is a fairly ideal place to put the tax token filter.

zavelevsky (approved):

fastlane_bot/bot.py
```diff
@@ -166,7 +166,10 @@ def get_curves(self) -> CPCContainer:
         for p in pools_and_tokens:
             try:
                 p.ADDRDEC = ADDRDEC
-                curves += p.to_cpc()
+                curves += [
```
sklbancor, line 169: didn't you add the TAXTOKENS property only to mainnnet config? wouldn't this fail
for other networks?

barakman: It is set to `[]` by default for all chains (then overridden specifically for ethereum/mainnet).

Issue #519, opened 2024-04-07
NETWORK_TENDERLY overrides the original NETWORK


barakman:
## Bug Description

In file `utils.py`, the following action overrides the original network configured by the user:
```
        if 'tenderly' in rpc_url:
            cfg.NETWORK = cfg.NETWORK_TENDERLY
```
Subsequently, there is no way to handle the execution in tenderly in the exact same way it is handled outside of tenderly.

For example, in file `TxHelpers.py` currently under PR #478:
```
        if self.cfg.NETWORK == self.cfg.NETWORK_ETHEREUM:
            self.use_access_list = True
            self.send_transaction = self._send_private_transaction
        else:
            self.use_access_list = False
            self.send_transaction = self._send_regular_transaction
```
The results of running on Ethereum are as followed:
```
+------------------+-------------+-------------+
|                  | Transaction | Access List |
+------------------+-------------+-------------|
| With Tenderly    | Regular     | Not applied |
+------------------+-------------+-------------+
| Without Tenderly | Private     | Applied     |
+------------------+-------------+-------------+
```
Private transactions cannot be used in Tenderly, so testing here is incomplete, but there's no other option.
But access lists CAN be used in Tenderly, so there IS an option here to improve the testing of Ethereum via Tenderly.

## Severity

Medium

## Solution Guidelines

Instead of `NETWORK_TENDERLY` being one of the values permitted for `Config.NETWORK`, the `Config` class should define an additional variable named `IS_TENDERLY` or similar.

This change might subsequently require a bunch of other changes across the codebase.

PR #529, opened 2024-04-09
Class TxHelpers Revision

barakman:
This PR includes several changes, mostly around coding simplification and performance improvements in the `TxHelpers` class, alongside a few related or partially-related cleanups in other modules.

The main changes in the `TxHelpers` class are listed below, and only two of them - 1 and 6 - are expected to impact the actual Arb-Behavior of the bot:

1. Replace the user-configurable `DEFAULT_GAS_PRICE_OFFSET` constant with a `gas_strategy` function.
2. Fetch the `maxPriorityFee` via web3 instead of by sending an HTTP request to the node.
3. Fetch the `gasPrice` instead of fetching the entire block in order to retrieve the `baseFee`.
4. Send a single gas-estimation request to the node.
5. Fetch the `accessList` via web3 instead of by sending an HTTP request to the node.
6. Remove the `EXPECTED_GAS_MODIFIER` constant used for producing an optimistic gas consumption.
7. Remove support for legacy-transactions; all chains are assumed to be supporting EIP-1559.
8. Send `eth_sendPrivateTransaction` HTTP requests via web3 instead of via the node.
9. Ensure that the access list is applied on Ethereum transactions; it is not applied in the current version due to a bug in the code, and even after this PR, it will not be tested within Tenderly (when the target chain is Ethereum) because of that same bug; see issue #519.
10. Construct the `tx` dictionary manually and then set its `gas` attribute by calling `estimate_gas`, instead of doing both of these things via `build_transaction`; following that, set the `accessList` attribute in this dictionary, and possibly update its `gas` attribute, by calling `create_access_list`; the gas-fee attributes (`maxFeePerGas` and `maxPriorityFeePerGas`) are set only at the end, in order to avoid getting the "max fee per gas less than block base fee" error.


Lesigh-3100:
Beyond code review, this PR should be live-tested - including normal & private transactions.

zavelevsky (approved):
code looks good to me, I didn't test it and I'd like more eyes other than mine

barakman:
Reposted via #542

PR #530, opened 2024-04-09
Update static data mapping files

Lesigh-3100:
Updated static data mapping files for:
-Ethereum
-Base
-Fantom
-Mantle

Note:
Large diff for some files - removed extra column that contained numbers for some pools in order to more effectively remove duplicates.


barakman (changes requested):
What about fixing the data-collection script to ensure no duplicated information?

zavelevsky (approved):

PR #531, opened 2024-04-10
Add support for XFAI on Linea

NIXBNT:
Added support for XFAI sufficient to populate the pool data correctly and arbs can be found (on Tenderly).
Unable to test terraformer due to rpc issues - so populated the static data manually.
Since, no RPC I couldn't test the event fetching so there may be still issues there.
Added notes on each commit description about improvements we should implement.
Note that the get_fee needs be handled correctly (using a different contract CORE) or alternatively hardcoded.
Executing arbs not tested as pending contract upgrade


run_blockchain_terraformer.py
```diff
@@ -147,6 +148,7 @@
     AERODROME_V2_NAME: 12,
     CLEOPATRA_V2_NAME: 12,
     STRATUM_V2_NAME: 12,
+    XFAI_NAME: 13 #TODO verify with contracts
```
barakman, line 151: Verified

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -173,3 +190,5 @@ async def get_tkn0(self, address: str, contract: Contract, event: Any) -> str:
     async def get_tkn1(self, address: str, contract: Contract, event: Any) -> str:
         return await contract.functions.token1().call()
 
+    async def get_xfai_tkn0(self, address: str, contract: Contract, event: Any) -> str:
+        return await contract.functions.poolToken().call()
```
barakman, line 194: Probably won't work, since function `poolToken` was not added in the ABI.

fastlane_bot/events/exchanges/solidly_v2.py
```diff
@@ -115,6 +115,22 @@ async def get_fee_6(address: str, contract: Contract, factory_contract: Contract
     """
     return await factory_contract.caller.pairFee(address)
 
+async def get_fee_7(address: str, contract: Contract, factory_contract: Contract) -> float:
```
barakman, line 118: TODO: pull from dev, then fix this to `_ get_fee_7... -> int`.

barakman (changes requested):
This DEX looks significantly different than all other Solidly V2 DEXs.
Consider separating it completely from that group.

fastlane_bot/events/exchanges/solidly_v2.py
NIXBNT: Even though the fee is unlikely to change we will need to fetch it from the CORE contract (i.e. this is different to the Factory contract) 
https://lineascan.build/address/0x8D58ee2D23f7920eA32e534aAD8d6753c88bc01A

NIXBNT:
I successfully executed arbs in a test environment.
Still requires a contract upgrade to function on mainnet

barakman:
Handled via #563

PR #534, opened 2024-04-10
chore: poetry project setup

platonfloria:
Added poetry package manager.
Removed conda, requirements.txt, setup.py and cleaned up CI.


sklbancor:
Can you tell us a bit more about poetry and why this change is a good idea at this point? Not doubting that it is a great system but I am slightly worried that it is a big breaking change for some of our users. 

Also this seems complicated: are we doing this?

> Poetry should always be installed in a dedicated virtual environment to isolate it from the rest of your system. It should in no case be installed in the environment of the project that is to be managed by Poetry. This ensures that Poetry’s own dependencies will not be accidentally upgraded or uninstalled. (Each of the following installation methods ensures that Poetry is installed into an isolated environment.) In addition, the isolated virtual environment in which poetry is installed should not be activated for running poetry commands.

https://python-poetry.org/docs/

pyproject.toml
```diff
@@ -0,0 +1,42 @@
+[tool.poetry]
+name = "fastlane-bot"
+version = "3.1.5"
+description = "Carbon Arbitrage Bot, an open-source arbitrage protocol, allows any user to perform arbitrage between Bancor ecosystem protocols and external exchanges and redirect arbitrage profits back to the protocol."
+authors = ["Bancor Network <platon@bancor.network>"]
```
sklbancor, line 5: author field should be our generic author address

pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+license = "MIT"
+readme = "README.md"
+packages = [{include = "fastlane_bot"}]
+
+[tool.poetry.dependencies]
+python = "^3.9"
```
sklbancor, line 11: thus far we said 3.8

pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+requests = "^2.31.0"
+python-dateutil = "^2.8.2"
+typing-extensions = "^4.7.1"
+python-dotenv = "^0.16.0"
+joblib = "^1.2.0"
+pandas = "^1.5.2"
+alchemy-sdk = "^0.1.1"
+pyarrow = "^11.0.0"
+networkx = "^3.0"
+cvxpy = "^1.3.1"
```
sklbancor, line 23: can be removed

sklbancor (changes requested):

pyproject.toml
platonfloria: This is the version that was present in setup.py

platonfloria:
As you can see from the changes that I've made, poetry very much simplifies our life, as it takes care of
- virtual environments (replacing conda)
- dependencies (replacing requirements.txt and pip)
- version bumping (instead of manual approach that was used previously)
- wheel build
- lockfile creation (which ensures that bot runs in the same environment across different deployments)

> Also this seems complicated: are we doing this? Poetry should always be installed in a dedicated virtual environment

This is all done behind the scenes by poetry, no additional action is required from us

pyproject.toml
platonfloria: done

platonfloria: Tests are failing without `cvxpy = "^1.3.1"`, adding it back

pyproject.toml
sklbancor: yeah. we currently test with 3.8. Maybe we should rethink this.

pyproject.toml
sklbancor: yeah. we also need to discuss this. those tests are for library code in the optimizer that is not in use in the bot proper.

README.md
```diff
@@ -178,15 +138,15 @@ You can configure the Fastlane Arbitrage Bot using the options in the `@click.op
...
+poetry run python main.py --arb_mode=multi --polling_interval=12 --reorg_delay=10 --loglevel=INFO
 ```
 
 ## Troubleshooting
 
 If you encounter import errors or `ModuleNotFound` exceptions, try:
 
 ```bash
-python <absolute_path>/main.py
+poetry run python <absolute_path>/main.py
```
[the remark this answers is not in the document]
barakman: Can or should we add that in the README file?

pyproject.toml
zavelevsky: services@bancor.network
Does this require opening an account somewhere?

pyproject.toml
platonfloria: I can move it to dev dependencies then, if anything else should be moved in there let me know

README.md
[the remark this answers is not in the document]
platonfloria: done

pyproject.toml
sklbancor: not sure that this is the ideal way of dealing with this -- maybe production users also want to run the test suite. Opening an issue.

sklbancor (approved):

barakman (approved):

PR #542, opened 2024-04-11
Class TxHelpers Revision (April 11)

barakman:
This PR includes several changes, mostly around coding simplification and performance improvements in the `TxHelpers` class, alongside a few related or partially-related cleanups in other modules.

The main changes in the `TxHelpers` class are listed below, and only two of them - 1 and 6 - are expected to impact the actual Arb-Behavior of the bot:

1. Replace the user-configurable `DEFAULT_GAS_PRICE_OFFSET` constant with a `gas_strategy` function.
2. Fetch the `maxPriorityFee` via web3 instead of by sending an HTTP request to the node.
3. Fetch the `gasPrice` instead of fetching the entire block in order to retrieve the `baseFee`.
4. Send a single gas-estimation request to the node.
5. Fetch the `accessList` via web3 instead of by sending an HTTP request to the node.
6. Remove the `EXPECTED_GAS_MODIFIER` constant used for producing an optimistic gas consumption.
7. Remove support for legacy-transactions; all chains are assumed to be supporting EIP-1559.
8. Send `eth_sendPrivateTransaction` HTTP requests via web3 instead of via the node.
9. Ensure that the access list is applied on Ethereum transactions; it is not applied in the current version due to a bug in the code, and even after this PR, it will not be tested within Tenderly (when the target chain is Ethereum) because of that same bug; see issue #519.
10. Construct the `tx` dictionary manually and then set its `gas` attribute by calling `estimate_gas`, instead of doing both of these things via `build_transaction`; following that, set the `accessList` attribute in this dictionary, and possibly update its `gas` attribute, by calling `create_access_list`; the gas-fee attributes (`maxFeePerGas` and `maxPriorityFeePerGas`) are set only at the end, in order to avoid getting the "max fee per gas less than block base fee" error.


fastlane_bot/helpers/txhelpers.py
```diff
@@ -133,579 +61,145 @@ def validate_and_submit_transaction(
...
+    def _send_regular_transaction(self, raw_tx: str) -> str:
+        return self.cfg.w3.eth.send_raw_transaction(raw_tx).hex()
 
-        :param method: the API method to call
-        """
-        response = requests.post(
-            self.alchemy_api_url,
-            json=self._get_payload(method=method, params=params),
-            headers=self._get_headers,
+    def _send_private_transaction(self, raw_tx: str) -> str:
```
Lesigh-3100, line 192: Couple points:

1. This must be tested & confirmed that it works. It may be difficult to find a profitable transaction on Ethereum, so a good testing option would be to force a transaction through that is not profitable after gas costs. 
2. This function likely only works with Alchemy. For someone using a different RPC or local node it would fail unless the RPC implements the `eth_sendPrivateTransaction` method - I'm not aware of non-Alchemy RPCs that have this aside from Flashbots themselves. I still believe implementing this call directly to Flashbots would be optimal as it would remove all reliance on Alchemy.

fastlane_bot/helpers/txhelpers.py
```diff
@@ -50,81 +39,20 @@ class TxHelpers:
...
-
-        # Gas cost in usd can be estimated using the profit usd/eth rate
-        gas_cost_usd = gas_cost_eth * expected_profit_usd / expected_profit_eth
-
-        # Multiply by reward percentage, taken from the arb contract
-        adjusted_reward_eth = Decimal(Decimal(expected_profit_eth) * Decimal(self.ConfigObj.ARB_REWARD_PERCENTAGE))
-        adjusted_reward_usd = adjusted_reward_eth * expected_profit_usd / expected_profit_eth
-
-        return gas_cost_eth, gas_cost_usd, adjusted_reward_eth, adjusted_reward_usd
+        self.chain_id = self.cfg.w3.eth.chain_id
```
Lesigh-3100, line 45: `chain_id` only needs to be called once per Bot run - can we move this to provider.py and fetch it internally from the `cfg` object?

Lesigh-3100 (changes requested):

fastlane_bot/helpers/txhelpers.py
barakman: 1. Yes - that has crossed my mind, with the main point in it being to make sure that the private-transaction mechanism is working correctly.
2. This is currently the case anyway, as far as I remember (the bot uses private transactions for Ethereum, regardless of which node is in use).

fastlane_bot/helpers/txhelpers.py
barakman: It IS called once, when the `TxHelpers` instance is created... well, maybe once more, for the token-approval part, but this is insignificant (and it needs to be fixed regardless of this PR, i.e., there only needs to be a single `TxHelpers` instance created for the entire system).

barakman: OK, this is even worse than I thought.
An instance of `CarbonBot` is recreated on every iteration of the main loop, and an instance of `TxHelpers` is recreated as part of it.
The part which needs to be fixed is this, i.e., creating the `CarbonBot` only once.
But of course - not in this PR.

zavelevsky (approved):

zavelevsky (approved):

Issue #543, opened 2024-04-11
Make the bot simple to run by setting appropriate defaults per chain


NIXBNT:
## Enhancement Description
- Summary of the enhancement
Switching the bot between different chains can be cumbersome and require chain specific knowledge of appropriate exchanges, flashloan tokens and min profit settings. 
By proposing set of default settings for each chain separately, the user interface can be significantly improved.
Running the chain should be as simple as inputting --blockchain=chain to the click options to get started, then users can optimize accordingly as their domain knowledge grows, rather than knowing the list of suitable flashloan tokens for example.

Solution: A list of click option defaults per chain

## Impact Analysis
- Automated Tests: Implications on testing (new/changed)
- Risks: What are the risks associated with this enhancement?
- Performance: Expected performance impacts and how to measure them
- Monitoring: Changes required in the monitoring system
- Logging: Do we need additional logs?
- Documentation: Updates or new documentation needed?

PR #544, opened 2024-04-12
Default click options per chain

NIXBNT:
Addresses #543 

Adds a set of default parameters for each supported blockchain to the constants.py

WARNING
1) These defaults need to be checked thoroughly prior to merge @barakman @Lesigh-3100 
2) In order to use uniswap_v2_forks for the default input of each exchange I was required to remove 'uniswap_v2' from being explicitly added to the forks list (as it doesnt appear on every chain). We need to verify this doesnt have any adverse effects @Lesigh-3100 thoughts? Alternatively, we can go back to specifying each supported exchange explicitly in the defaults


fastlane_bot/data/multichain_addresses.csv
```diff
@@ -49,7 +49,6 @@ sushiswap_v3,core,uniswap_v3,0xc35DADB65012eC5796536bD9864eD8773aBc74C4,N/A,,,
 fraxswap_v2,dogechain,uniswap_v2,0x67b7DA7c0564c6aC080f0A6D9fB4675e52E6bF1d,0x0f6A5c5F341791e897eB1FB8fE8B4e30EC4F9bDf,,,
 balancer,ethereum,balancer,BALANCER_VAULT_ADDRESS,0xBA12222222228d8Ba445958a75a0704d566BF2C8,0,0,
 carbon_v1,ethereum,carbon_v1,0xC537e898CD774e2dCBa3B14Ea6f34C93d5eA45e1,0xC537e898CD774e2dCBa3B14Ea6f34C93d5eA45e1,NA,17087375,
-altered_carbon,ethereum,carbon_v1,0xA31ac5a9077d8EE58F4cd0011C315552De29542B,0xA31ac5a9077d8EE58F4cd0011C315552De29542B,NA,19375311,
```
barakman, line 52: How is this change related here?

barakman (changes requested):

fastlane_bot/data/multichain_addresses.csv
NIXBNT: not a supported exchange so it breaks the use of carbon_v1_forks on ethereum - should be deleted

PR #546, opened 2024-04-14
Fix the np exp overflow issue

barakman:
**CODE IS NO LONGER IN THIS REPO; REOPEN THE ISSUE IN THE OPTIMIZER REPO**


sklbancor (changes requested):
let's create some tests that show what this solves

barakman:
> let's create some tests that show what this solves

We already have one - file `test_901_TestMultiTriangleModeSlow.py` function `test_test_combos`.

PR #563, opened 2024-04-21
Add support for XfaiV0 as a fork of SolidlyV2

barakman:
Based on PR #531


barakman:
Data files not yet updated

fastlane_bot/data/abi.py
```diff
@@ -286,7 +286,31 @@
...
+XFAI_V0_FACTORY_ABI = [
+    {
+        "type": "event",
+        "name": "PoolCreated",
+        "anonymous": False,
+        "inputs": [{"indexed": True, "internalType": "address", "name": "token", "type": "address"}, {"indexed": True, "internalType": "address", "name": "pool", "type": "address"}, {"indexed": False, "internalType": "uint256", "name": "allPoolsSize", "type": "uint256"}]
+    },
```
barakman, line 295: @NIXBNT @Lesigh-3100 :
This event prototype is not the same as that of SolidlyV2.
How can this be handled correctly?

barakman: I think that each one of the following lines in function `organize_pool_details_solidly_v2` in the terraformer script would throw an exception:
- `stable_pool = "stable" if pool_data["args"]["stable"] else "volatile"`
- `tokens = [pool_data["args"]["token0"], pool_data["args"]["token1"]]`

This function probably needs to be extended specifically for XfaiV0.

NIXBNT (approved):

zavelevsky (approved):

Issue #556, opened 2024-04-19
Flashbots does not seem to handle access lists


barakman:
## Enhancement Description
- Flashbots private transactions do not seem to support access lists
- This yields an increased gas cost for arbitrage transactions
- Check the option of switching to Mevblocker

PR #569, opened 2024-04-24
Refrain from including the access-list when it contains an error

barakman:
Partially handles issue #556:
- Still need to verify that it works with flashbots
- Still need to check if it works with mevblocker
- Still need to re-enable `self.use_access_list`

### DIFF
--- a/fastlane_bot/helpers/txhelpers.py
+++ b/fastlane_bot/helpers/txhelpers.py
@@ -168,10 +168,10 @@ class TxHelpers:
         }
 
     def _update_transaction(self, tx: dict):
-        tx["gas"] = self.cfg.w3.eth.estimate_gas(tx) # occasionally throws an exception
+        tx["gas"] = self.cfg.w3.eth.estimate_gas(tx) # may throw an exception
         if self.use_access_list:
-            result = self.cfg.w3.eth.create_access_list(tx) # rarely throws an exception
-            if tx["gas"] > result["gasUsed"]:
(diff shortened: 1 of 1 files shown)


fastlane_bot/helpers/txhelpers.py
```diff
@@ -168,10 +168,10 @@ def _create_transaction(self, contract, fn_name: str, args: list, value: int) ->
         }
 
     def _update_transaction(self, tx: dict):
-        tx["gas"] = self.cfg.w3.eth.estimate_gas(tx) # occasionally throws an exception
+        tx["gas"] = self.cfg.w3.eth.estimate_gas(tx) # may throw an exception
         if self.use_access_list:
-            result = self.cfg.w3.eth.create_access_list(tx) # rarely throws an exception
-            if tx["gas"] > result["gasUsed"]:
+            result = self.cfg.w3.eth.create_access_list(tx) # may return an error
+            if tx["gas"] > result["gasUsed"] and "error" not in result:
```
zavelevsky, line 174: I'd check the result is valid before accessing specific fields in it

zavelevsky (changes requested):

fastlane_bot/helpers/txhelpers.py
barakman: It is valid by the definition of function `create_access_list` - either it returns the `result` object, or it throws an exception (for example, if there is no network connection at that given point in time).

zavelevsky (approved):

PR #573, opened 2024-04-24
Update data files 24 april

barakman:


fastlane_bot/data/blockchain_data/fantom/tokens.csv
```diff
@@ -18465,8 +18465,8 @@ address,decimals,symbol
 0x1f82e3d045A7D16deE6428E8f0c87bCf75CDF685,9,HYPER
 0xf26078B830AB1Eb13B5B5b35044E2dE96D67838C,18,Matrix
 0x6228Fbd41A91A441976f11907e78A2A6b41BEe54,9,tertes
-0x63DDb80DD021734b7C7EB44123278C070a4f73dB,9,NA
-0xB3B56B2b2F5be49573bBeD277Cf2eD9d02236629,9,NA
+0x63DDb80DD021734b7C7EB44123278C070a4f73dB,9,
```
zavelevsky, line 18468: is it legit to just remove the NA?

zavelevsky (dismissed):

fastlane_bot/data/blockchain_data/fantom/tokens.csv
barakman: I was wondering about this myself, but that's what the terraformer script produced, so I couldn't argue with it.

barakman: In either case, data files for both ethereum and fantom still need to be updated, so it might be worth visiting this comment afterwards (I'm not really sure what exactly caused this file to get updated).

barakman: Yes, it appears that these token symbols have been added manually (otherwise, there should indeed be a problem when attempting to read this CSV file via Panda).
I have fixed the terraformer script to add the token symbol `???` automatically when needed.
This file will be fixed by running that script for Fantom and pushing the updated data files.

zavelevsky (approved):

PR #575, opened 2024-04-25
Chore/update ci

platonfloria:
Updated CI to generate requirements.txt


zavelevsky (approved):

zavelevsky (changes requested):
this action only runs when merging into main. We are running databricks for branches - so we will have branches without requirements.txt

zavelevsky:
unneeded any more

PR #578, opened 2024-04-25
Test900 fails

sklbancor:
this comments out the convex optimizer code that is not used in prod and that seems to be unreliable when testing on github

### DIFF
--- a/resources/NBTest/NBTest_900_OptimizerDetailedSlow.py
+++ b/resources/NBTest/NBTest_900_OptimizerDetailedSlow.py
@@ -444,6 +444,9 @@ assert ti0.tknin == f"{T.USDT}"
 assert ti0.tknout == f"{T.USDC}"
 assert round(ti0.amtin, 8)  == 1214.45596849
 assert round(ti0.amtout, 8) == -1216.41933959
+if not ti0.error is None:
+    print(ti0)
+    print(ti0.error)
 assert ti0.error is None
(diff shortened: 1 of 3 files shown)


resources/NBTest/NBTest_900_OptimizerDetailedSlow.py
```diff
@@ -444,6 +444,9 @@
 assert ti0.tknout == f"{T.USDC}"
 assert round(ti0.amtin, 8)  == 1214.45596849
 assert round(ti0.amtout, 8) == -1216.41933959
+if not ti0.error is None:
+    print(ti0)
+    print(ti0.error)
 assert ti0.error is None
```
barakman, line 450: How about `assert ti0.error is None, f'{ti0.error}'` instead of those two `print` statements?

barakman (approved):
One minor comment for you to decide on; other than that - approved.

resources/NBTest/NBTest_900_OptimizerDetailedSlow.py
sklbancor: it's somewhat nicer and cleaner but I actually want to print the whole object, not only the error

barakman: So `assert ti0.error is None, f'{ti0}'`...

PR #582, opened 2024-04-29
DRAFT: Add websocket usage

platonfloria:


fastlane_bot/events/utils.py
```diff
@@ -1284,14 +1271,12 @@ def get_latest_events(
...
-        complex_handler(event)
-        for event in [
-            complex_handler(event)
-            for event in get_all_events(
-                n_jobs,
-                get_event_filters(n_jobs, mgr, start_block, current_block),
-            )
-        ]
+        Event.from_dict(complex_handler(event))
+        for events in get_all_events(
```
sklbancor, line 1275: minor point -- but having the loop variable and the final output variable both called `events` is slightly confusing

fastlane_bot/events/listener.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    NEW_EVENT_TIMEOUT = 0.01
+
+    def __init__(
+        self,
+        manager: BaseManager,
+        base_exchanges: List[str],
+        w3: AsyncWeb3,
+    ):
+        """ Initializes the EventManager.
```
sklbancor, line 24: I tend to attack the init doc string to the class because of the call semantics

obj = MyClass(a,b,c)

as opposed to 

obj = MyClass.__init__(a,b,c)

fastlane_bot/events/listener.py
```diff
@@ -0,0 +1,72 @@
...
+        self._w3 = w3
+        self._subscriptions = []  # Initialize the subscription list
+        self._subscription_by_id = {}
+        self._event_buffer = []
+        self._last_event_ts = 0
+
+        for exchange_name, exchange in self._manager.exchanges.items():
+            if exchange_name in base_exchanges:
+                self._subscriptions += exchange.get_subscriptions(self._manager.event_contracts[exchange_name])
+
```
sklbancor, line 40: isn't this just essentially
```
self._subscriptions = [
exchange.get_subscriptions(self._manager.event_contracts[exchange_name]
for exchange_name in set(self._manager.exchanges).intersect(base_exchanges)
]
```

sklbancor (commented):
just some minor things I noticed when looking through the code.

I did not really find what I was looking for unfortunately, which is the global flow that (a) spawns the event log, and (b) calls the optimizer and thereafter the execution

fastlane_bot/events/listener.py
platonfloria: You would need to flatten it as well

fastlane_bot/events/exchanges/bancor_v2.py
```diff
@@ -59,15 +65,15 @@ async def get_fee(self, address: str, contract: AsyncContract) -> Tuple[str, flo
...
+            return event.args["_token1"]
+        return await contract.functions.reserveTokens()[0]
 
-    async def get_tkn1(self, address: str, contract: Contract, event: Any) -> str:
+    async def get_tkn1(self, address: str, contract: Contract, event: Event) -> str:
         if event:
-            return event["args"]["_token2"]
-        return await contract.caller.reserveTokens()[1]
+            return event.args["_token2"]
+        return await contract.functions.reserveTokens()[1]
```
barakman, line 76: Should be `caller`, not `functions`.

fastlane_bot/events/listener.py
barakman: It's equivalent (without the need for an additional flattening) to:
```
self._subscriptions = [
    exchange.get_subscriptions(self._manager.event_contracts[exchange_name])
    for exchange_name, exchange in self._manager.exchanges.items()
    if exchange_name in base_exchanges
]
```

fastlane_bot/events/exchanges/bancor_v2.py
```diff
@@ -59,15 +65,15 @@ async def get_fee(self, address: str, contract: AsyncContract) -> Tuple[str, flo
             fee_float = float(fee) / 1e6
         return fee, fee_float
 
-    async def get_tkn0(self, address: str, contract: Contract, event: Any) -> str:
+    async def get_tkn0(self, address: str, contract: Contract, event: Event) -> str:
         if event:
-            return event["args"]["_token1"]
-        return await contract.caller.reserveTokens()[0]
+            return event.args["_token1"]
+        return await contract.functions.reserveTokens()[0]
```
barakman, line 71: Should be `caller`, not `functions`.

barakman (changes requested):

PR #253, opened 2023-12-12
Convergence issues - SUPERSEDED BY #588

sklbancor:
**Superseded by #588; all changes have now been moved over; will be closed**

Once #588 is merged this one can be closed. Do not make changes in this branch. It should forever remain at https://github.com/bancorprotocol/fastlane-bot/commit/c453c8362e0f30778f9892f6f49b2e7f9da90ec3

This pull request is dealing with the convergence issues we experienced with big limit orders. 

Major changes:
- CPC v3.3.3: as_df, as_dicts (addressed inconsistent naming; old function deprecated but remain)
- CPC v3.4: added minrw parameter to CPC.from_carbon to allow for different choices of the minimum size of the range

Also added a resources/testing directory; the Jupyter notebook therein is the main source of our research on this topic that underlies those improvement; it also contains unit tests that should be moved to the NBTest area (**TODO BEFORE MERGE**)

PR #588, opened 2024-04-30
Optimizer Convergence 2

sklbancor:
**THIS BRANCH IS NOW FINALIZED; FUTURE WORK IN THE CONVERSION 3 #602 BRANCH**

note: this branch is meant to be on top of #606 

This PR is dealing with the convergence issues with found. It builds on #253 where those issues had been addressed by offering an alternative _absolute convergence_ test. Convergence has been assured but we haven't had a chance of fine tuning the parameters yet. One issue that it was designed to deal with (the one in the `Optimizer)_2405_SEI` sheets) turns out can at this stage not effectively be addressed at the Optimizer level. However, at the bot level a quick fix exists and #594  has been created. Long run, this should however bee addressed in the Optimizer, and issue #595 has been raised

We are also looking at a second convergence issue that we have come across in relation to the upcoming SEI release, which is why this one is on top of the `add_sei` branch rather than on top of `develop`.

### Notes from #253 

Major changes:
- CPC v3.3.3: as_df, as_dicts (addressed inconsistent naming; old function deprecated but remain)
- CPC v3.4: added minrw parameter to CPC.from_carbon to allow for different choices of the minimum size of the range


sklbancor:
```
=========================== short test summary info ============================
FAILED fastlane_bot/tests/nbtest/test_002_CPCandOptimizer.py::test_margp_optimizer
FAILED fastlane_bot/tests/nbtest/test_003_Serialization.py::test_creating_curves
============ 2 failed, ***77 passed, 33 warnings in 547.34s (0:09:07) ============
```

fastlane_bot/tools/optimizer/margpoptimizer.py
```diff
@@ -236,14 +248,15 @@ def optimize(self, sfc=None, result=None, *, params=None):
...
+                    verbose=P("debug_pe"),
                     triangulate=True,
                 )
             except Exception as e:
                 if P("verbose") or P("debug"):
                     print(f"[margp_optimizer] error calculating price estimates: [{e}]")
                 price_estimates_t = None
-                raise
+                if not result in [self.MO_PSTART, self.MO_DEBUG]:
+                    raise
```
barakman, line 259: Consider something more informative, for example:
```
assert result in [self.MO_PSTART, self.MO_DEBUG], f"{result}"
```

barakman (approved):
File `resources/docs/202312 ArbBot Convergence.pdf` should probably be removed from the repository.

BRANCH_CONVERGENCE2
```diff
@@ -0,0 +1,9 @@
+TESTS PASSED (LATEST FIRST)
+----------------------------------------
+8ab28875bf04d679b31b98e960458d9c0d77e5c5
+942ccee629987c9496e1e03071c0ac0933c16a54
+562c88ce29a29dbab52b1513ebb8343ed73ee0f0
+e892b62036fe5368b9b617f8d21a9dc49bee08f6
+ccdf8617c7b93adcf1fc57e741262140744a66c3
+
+
```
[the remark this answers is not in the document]
sklbancor: 1/ it shows me where exactly the topic branch starts which allows me to do some black git magic and recover from almost any rebase fuckup that happens in the repo
2/ it contains hashes of configurations that passed the tests, so I know which is the last known configuration

Any trace of this file will be removed before merging.

fastlane_bot/tools/optimizer/margpoptimizer.py
sklbancor: not the same; `raise` re-raises the exception, `assert` raise a different exception
also this is actually not the point here -- I want the code to continue in those cases (which are non-standard execution paths) because they'll return early anyway

PR #593, opened 2024-05-01
Optimizer5.3 (partial commit of convergence 2)

sklbancor:
This is an early commit of settled features within the #588 and #253 pull requests. It introduces potentially significant changes under the hood. I do not expect things to fail -- and tests don't -- but this is big enough to warrant a certain period of testing

Major changes
- Optimizer now allows for absolute convergence checks; those improve convergence but requires USD rates; by default they are disabled
- CPC.from_carbon  on constant price curves now produces slightly more convex curves for better numeric stability; **this can have knock on effects**
- there is a trivial change to SimplePair that should not affect anything

### DIFF
--- a/fastlane_bot/tools/cpc.py
+++ b/fastlane_bot/tools/cpc.py
@@ -8,8 +8,8 @@ Licensed under MIT
 NOTE: this class is not part of the API of the Carbon protocol, and you must expect breaking
 changes even in minor version updates. Use at your own risk.
 """
-__VERSION__ = "3.4"
-__DATE__ = "23/Jan/2024"
+__VERSION__ = "3.5"
+__DATE__ = "22/Apr/2023"
 
 from dataclasses import dataclass, field, asdict, InitVar
 from .simplepair import SimplePair as Pair
@@ -836,7 +836,7 @@ class ConstantProductCurve(CurveBase):
             constr="uv3",
             params=params,
         )
-
+    
     SOLIDLY_PRICE_SPREAD = 0.06     # 0.06 gives pretty good results for m=2.6
     @classmethod
     def from_solidly(
@@ -982,6 +982,9 @@ class ConstantProductCurve(CurveBase):
             print("[cpc::from_solidly] returning curve directly is deprecated; prepare to accept a list of curves in the future")
             return result
         
+    # minimun range width (pa/pb-1)  for carbon curves and sqrt thereof
+    CARBON_MIN_RANGEWIDTH  = 1e-6 
+    
     @classmethod
     def from_carbon(
         cls,
@@ -999,6 +1002,7 @@ class ConstantProductCurve(CurveBase):
         descr=None,
         params=None,
         isdydx=True,
+        minrw=None,
     ):
         """
         constructor: from a single Carbon order (see class docstring for other parameters) (1)
@@ -1011,6 +1015,7 @@ class ConstantProductCurve(CurveBase):
         :B:         alternative to pa, pb: B = sqrt(pb) in dy/dy
         :tkny:      token y
         :isdydx:    if True prices in dy/dx, if False in quote direction of the pair
+        :minrw:     minimum perc width (pa/pb-1) of range (default CARBON_MIN_RANGEWIDTH)
 
         NOTE 1: that ALL parameters are mandatory, except that EITHER pa, bp OR A, B
         must be given but not both; we do not correct for incorrect assignment of
@@ -1028,7 +1033,10 @@ class ConstantProductCurve(CurveBase):
         # assert not fee is None, "fee must not be None"
         # assert not cid is None, "cid must not be None"
         # assert not descr is None, "descr must not be None"
-
+        
+        if minrw is None:
+            minrw = cls.CARBON_MIN_RANGEWIDTH
+        
         # if yint is None:
         #     yint = y
         assert y <= yint, "y must be <= yint"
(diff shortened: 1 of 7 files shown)


fastlane_bot/tools/cpc.py
```diff
@@ -1067,15 +1075,21 @@ def from_carbon(
...
+                # pb /= 1.0000001
+                rw_multiplier = sqrt(1+minrw)
+                pa *= rw_multiplier
+                pb /= rw_multiplier
 
             # validation
-            if not pa > pb:
-                raise cls.CPCValidationError(f"pa > pb required ({pa}, {pb})")
+            if not pa/pb - 1 >= minrw*0.99:
+                raise cls.CPCValidationError(f"pa +> pb required ({pa}, {pb}, {pa/pb-1}, {minrw})")
```
barakman, line 1092: The `+` in the error-message seems like a typo.

fastlane_bot/tools/cpc.py
```diff
@@ -1067,15 +1075,21 @@ def from_carbon(
...
+                # pa *= 1.0000001
+                # pb /= 1.0000001
+                rw_multiplier = sqrt(1+minrw)
+                pa *= rw_multiplier
+                pb /= rw_multiplier
 
             # validation
-            if not pa > pb:
-                raise cls.CPCValidationError(f"pa > pb required ({pa}, {pb})")
+            if not pa/pb - 1 >= minrw*0.99:
```
barakman, line 1091: I'm pretty sure that you should use `abs(pa/pb - 1)` here.

fastlane_bot/tools/optimizer/margpoptimizer.py
```diff
@@ -155,247 +187,288 @@ def optimize(self, sfc=None, result=None, *, params=None):
...
+                print( "1/p   ", ", ".join(f"{1/x:,.2f}" for x in p))
+            
+            # pvec is dict {tkn -> (log) price} for all tokens in p
+            pvec = {tkn: p_ for tkn, p_ in zip(tokens_t, p)}
+            pvec[targettkn] = 1
+            
+            sum_by_tkn = {t: 0 for t in alltokens_s}
+            for pair, (tknb, tknq) in zip(pairs, pairs_t):
+                if get(p, tokens_ix.get(tknq)) > 0:
+                    price = get(p, tokens_ix.get(tknb)) / get(p, tokens_ix.get(tknq))
```
barakman, line 308: You might want to `get(p, tokens_ix.get(tknq))` instead of doing this computation twice.

fastlane_bot/tools/optimizer/margpoptimizer.py
```diff
@@ -155,247 +187,288 @@ def optimize(self, sfc=None, result=None, *, params=None):
...
+                else:
+                    #print(f"[dtknfromp_f] warning: price for {pair} is unknown, using 1 instead")
+                    price = 1
```
barakman, line 311: What if `get(p, tokens_ix.get(tknb)) > 0 and get(p, tokens_ix.get(tknq)) == 0`?
This doesn't really fall under the category of "unknown price", does it?
At least mathematically, it should yield an infinite price.
Consider raising an exception in this case.

fastlane_bot/tools/optimizer/margpoptimizer.py
```diff
@@ -155,247 +187,288 @@ def optimize(self, sfc=None, result=None, *, params=None):
...
+
+            :p:         price vector, where prices use the reference token as quote token
+                        this vector is an np.array, and the token order is the same as in tokens_t
+            :islog10:   if True, p is interpreted as log10(p)
+            :asdct:     if True, the result is returned as dict AND tuple, otherwise as np.array
+            :quiet:     if overrides P("debug") etc, eg for calc of Jacobian
+            :returns:   if asdct is False, a tuple of the same length as tokens_t detailing the
+                        change in token amounts for each token except for the target token (ie the
+                        quantity with target zero; if asdct is True, that same information is
+                        returned as dict, including the target token.
```
barakman, line 289: FYI, a function returning different output types under different circumstances, requires every caller of that function to be aware of this and handle the output accordingly.
Consider splitting this function into several smaller functions, each of which returning a single (consistent) type of output, which will relieve the caller or callers from having to do that sort of special handling.

fastlane_bot/tools/optimizer/margpoptimizer.py
```diff
@@ -155,247 +187,288 @@ def optimize(self, sfc=None, result=None, *, params=None):
...
+                # nrm = normf(dplog10)
+                # if nrm > 0.1:
+                #     dplog10 /= nrm
+                
+                # #### END EXPERIMENTAL
+                
+                # update log prices, prices...
+                p0log10 = [*plog10]                 # keep current log prices (deep copy)
+                plog10 += dplog10                   # update log prices
+                p = np.exp(plog10 * np.log(10))     # expand log to actual prices
```
barakman, line 428: If this piece of code is used excessively, then calculating the value of `np.log(10)` in advance and using it as a constant may serve for better performance.

barakman (changes requested):

fastlane_bot/tools/cpc.py
sklbancor: no; as you see I am testing here for pa > pb; I am just testing with a small tolerance, eg if the ranges have been created in a loop

fastlane_bot/tools/optimizer/margpoptimizer.py
sklbancor: yeah; this always sucks: speed vs garbage collection; more interesting though -- when looking at the code I realized that none of this was needed (it was only produced for logging and logging here is currently commented out) so I removed it...

fastlane_bot/tools/optimizer/margpoptimizer.py
sklbancor: that's the whole point of the if statement just above, no? am I missing something? anyway -- the code is removed

fastlane_bot/tools/optimizer/margpoptimizer.py
sklbancor: yeah; was thinking about this as well; I tend to avoid numeric optimization on a gut feel basis those days: there is not need to make code harder to read and more error prone for a 0.01% improvement in performance. I may have mentioned a few (hundred) times that I really really would love to profile that code in a live context

fastlane_bot/tools/optimizer/margpoptimizer.py
barakman: As I understand this, the paradigm should be:
1. If `n > 0 and d > 0` then return `price = n / d`
2. If `n = 0 and d = 0` then return `price = 1`
3. If `n > 0 and d = 0` then return `price = inf` or throw an exception

In your implementation, you are handling case 3 as if it was case 2.

sklbancor: that was exactly the plan; keep in mind this is only for debug output anyway; I used to have it without if, this crashed on me for no good reason [ie I did not want it to crash] so I added the if; needed to set it to a number because I use formatting, so I set it to 1. could have set it to zero I guess

barakman (approved):

PR #594, opened 2024-05-02
Convergence issues with overlapping strategies

sklbancor:
**MUST BE MERGE INTO #606 ASAP**

Overlapping strategies are designed to recreate Uniswap v3 like strategies on Carbon, ie ranges that trade bidirectional taking a small fee. Unsurprisingly -- in hindsight -- those strategies behave as badly within the Optimizer as other strategies with fees would (see #588). There is a reason that we ignore fees in the Optimizer and add them back only in the fine tuning step. 

High level the solution is actually pretty easy: **remove the effective fee component from overlapping strategies**. The current PR is designed to achieve that. However, it is a bit hacky, and once we have time we need to also take into account a few considerations spelled out in the next comment

### DIFF
--- a/fastlane_bot/helpers/poolandtokens.py
+++ b/fastlane_bot/helpers/poolandtokens.py
@@ -403,6 +403,14 @@ class PoolAndTokens:
         allow to omit yint (in which case it is set to y, but this does not make
         a difference for the result)
         """
+        def calculate_parameters(y: Decimal, pa: Decimal, pb: Decimal, pm: Decimal, n: Decimal):
+            H = pa.sqrt() ** n
+            L = pb.sqrt() ** n
+            M = pm.sqrt() ** n
+            A = H - L
+            B = L
+            z = y * (H - L) / (M - L) if M > L else y
+            return z
 
         # if idx == 0, use the first curve, otherwise use the second curve. change the numerical values to Decimal
         lst = []
@@ -523,14 +531,14 @@ class PoolAndTokens:
                 
                 # modify the y_int based on the new geomean to the limit of y    #TODO check that this math is correct
                 typed_args0 = strategy_typed_args[0]
-                new_yint0 = typed_args0['y'] * (typed_args0['pa'] - typed_args0['pb']) / ((1/geomean_p_marg) - typed_args0['pb'])  #this geomean is flipped because we flippend the pmarg from order 0
+                new_yint0 = calculate_parameters(y=typed_args0['y'], pa=typed_args0['pa'], pb=typed_args0['pb'], pm=(1/geomean_p_marg), n=1)
                 if new_yint0 < typed_args0['y']:
                     new_yint0 = typed_args0['y']
-                self.ConfigObj.logger.debug(f"[poolandtokens.py, _carbon_to_cpc] First order: typed_args0['yint'], new_yint0, , typed_args0['y']: {typed_args0['yint'], new_yint0, typed_args0['y']}")
+                self.ConfigObj.logger.debug(f"[poolandtokens.py, _carbon_to_cpc] First order: typed_args0['yint'], new_yint0, typed_args0['y']: {typed_args0['yint'], new_yint0, typed_args0['y']}")
                 typed_args0['yint'] = new_yint0
 
                 typed_args1 = strategy_typed_args[1]
-                new_yint1 = typed_args1['y'] * (typed_args1['pa'] - typed_args1['pb']) / (geomean_p_marg - typed_args1['pb'])
+                new_yint1 = calculate_parameters(y=typed_args1['y'], pa=typed_args1['pa'], pb=typed_args1['pb'], pm=(geomean_p_marg), n=1)
                 if new_yint1 < typed_args1['y']:
                     new_yint1 = typed_args1['y']
                 self.ConfigObj.logger.debug(f"[poolandtokens.py, _carbon_to_cpc] Second order: typed_args1['yint'], new_yint1, typed_args1['y']: {typed_args1['yint'], new_yint1, typed_args1['y']} \n")
--- a/fastlane_bot/helpers/poolandtokens.py
+++ b/fastlane_bot/helpers/poolandtokens.py
@@ -407,6 +407,7 @@ class PoolAndTokens:
         # if idx == 0, use the first curve, otherwise use the second curve. change the numerical values to Decimal
         lst = []
         errors = []
+        strategy_typed_args = []
         for i in [0, 1]:
 
             S = Decimal(self.A_1) if i == 0 else Decimal(self.A_0)
@@ -450,6 +451,7 @@ class PoolAndTokens:
             decimal_converter = decimal_converter(i)
 
             p_start = Decimal(encoded_order.p_start) * decimal_converter
+            p_marg = Decimal(encoded_order.p_marg) * decimal_converter
             p_end = Decimal(encoded_order.p_end) * decimal_converter
             yint = Decimal(yint) / (
                 Decimal("10") ** [self.tkn1_decimals, self.tkn0_decimals][i]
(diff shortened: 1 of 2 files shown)


sklbancor:
The current strategy dealing with effective fee removal for overlapping strategies works, but there are some issues that need to be address in the long run

- The optimal solution depends on the size of the fee (or bid/offer; the two are the same): the smaller the fee, the worse the Optimizer is affected, and the less impact on the overall solution it has to remove it; the bigger the fee, the more benign it is for the Optimizer (eg, 10% fee is definitely not a problem) and the worse it gets when we remove it; so we need to work on the threshold between removing the fee or not
- Where those adjustments are to be made is not clear; a priori that is a numerical issue that is better understood on the Optimizer / CPC side (and definitely not on the Node side); however, there are admin issues to consider as well, eg regarding CIDs
- In the long run, we should probably provide a Carbon strategy constructor that encapsulates the current Carbon order constructor because then the CPC has the information it needs to understand whether or not to make adjustments (at the moment it just sees different orders and it does not connect them); we discussed the issues around CIDs of this approach with Nick, and possibly the move to a CID / Sub CID approach
- Relatedly we noticed that the current CPC code creates curves that are optimized for the MargPOptimizer (notably, enforcing minimum curvature; the new code modifying strategies in place would fall into the same category); whilst this kinda makes sense -- we really only use MargPOptimizer -- this is not clean: the curves should be created as they are, and modified when reading into the Optimizer; however, this is a major change and it may not in practice be worth the effort.

fastlane_bot/helpers/poolandtokens.py
```diff
@@ -476,6 +485,67 @@ def decimal_converter(idx):
...
+            pmarg0, pmarg1 = [x['p_marg'] for x in strategy_typed_args]
+            pmarg0_inv = 1/pmarg0  # one of the orders must be flipped since prices are always dy/dx - but must flip same geomean_pmarg later
+            percent_component = pmarg_threshold * max(pmarg0_inv, pmarg1)
+            percent_component_met = abs(pmarg0_inv - pmarg1) <= percent_component
+
+            # overlapping strategies by defintion cannot have A=0 i.e. there must be no limit orders
+            no_limit_orders = (strategy_typed_args[0]['is_limit_order'] == False) and (strategy_typed_args[1]['is_limit_order'] == False)
+
+            # evaluate if the price boundaries pa/pb overlap at one end # TODO check logic and remove duplicate logic if necessary
+            prices_overlap = (strategy_typed_args[1]['pa']>(1/strategy_typed_args[0]['pa'])>strategy_typed_args[1]['pb']) # or (1/strategy_typed_args[0]['pa']<(strategy_typed_args[1]['pb']))
```
barakman, line 508: I think that it should be:
```
prices_overlap = max(Pb0, 1 / Pb1) < min(Pa0, 1 / Pa1)
```

barakman (changes requested):

fastlane_bot/helpers/poolandtokens.py
```diff
@@ -476,6 +485,67 @@ def decimal_converter(idx):
                 "descr": self.descr,
                 "params": self._params,
             }
+
+            strategy_typed_args += [typed_args]
+
+        #### MODIFICATION LOGIC HERE >>>>>
+        # Only overlapping strategies are selected for modification
+        if len(strategy_typed_args) == 2:
```
barakman, line 493: Unrelated to the changes in this PR, but each one of the two attempts to "try / except decimal.InvalidOperation" is completely redundant, because if the code prior to this attempt does not throw any exception, then so will the code inside this attempt.

barakman (changes requested):

fastlane_bot/helpers/poolandtokens.py
```diff
@@ -476,6 +485,67 @@ def decimal_converter(idx):
...
+            
+            is_overlapping = False
+            pmarg_threshold = Decimal("0.01") # 1%  # WARNING using this condition alone can included stable/stable pairs incidently
+
+            # evaluate that the marginal prices are within the pmarg_threshold
+            pmarg0, pmarg1 = [x['p_marg'] for x in strategy_typed_args]
+            pmarg0_inv = 1/pmarg0  # one of the orders must be flipped since prices are always dy/dx - but must flip same geomean_pmarg later
+            percent_component = pmarg_threshold * max(pmarg0_inv, pmarg1)
```
barakman, line 501: It feels like instead of using `max(pmarg0_inv, pmarg1)`, you should use either one of these two values, based on the direction in which you are measuring the relative difference, which could lead to different results.

For example:
- `percent_component_met = abs(100 - 99) <= 0.01 * max(100, 99)` --> `True`
- `percent_component_met = abs(100 - 99) <= 0.01 * min(100, 99)` --> `False`

Of course, this means that the current method needs to be aware of the arb in context (i.e., it cannot be oblivious to the direction of the arbitrage being inspected).

I'm not sure if this makes sense, so FYI.

barakman (changes requested):

sklbancor:
@NIXBNT just making sure we agree that this is your PR; I just opened it for you

fastlane_bot/helpers/poolandtokens.py
NIXBNT: Good suggestion but we wont implement for two reasons:
1) At this stage in the process we have no way of knowing the direction of the arb
2) This threshold is somewhat arbitrary and approximate solutions are ok. We expect the issue to occur around the 1% mark not necessarily a hard boundary and further investigation is ongoing

fastlane_bot/helpers/poolandtokens.py
NIXBNT: yeah I like this better too. Adding

fastlane_bot/utils.py
```diff
@@ -179,15 +183,27 @@ def p_start(self):
...
+            new_method = self.decodeRate(self.decodeFloat(int(self.B)) + self.decodeFloat(int(self.A)))
+            # except:
+            #     print(self.B, self.A)
+            #     print(type(self.B), type(self.A))
+            #     print(self.decodeFloat(int(self.B)) + self.decodeFloat(int(self.A)))
+            # print(new_method, self.p_start)
+            assert new_method == self.p_start, f"{new_method}, {self.p_start} **************************************"
             return self.p_start
         elif self.y == 0:
             return self.p_end
```
barakman, line 196: You can get rid of this `elif` part, since it is handled correctly in the `else` part.

fastlane_bot/utils.py
```diff
@@ -179,15 +183,27 @@ def p_start(self):
...
-        raise NotImplementedError("p_marg not implemented for non-full / empty orders")
-        A = self.decodeFloat(self.A)
-        B = self.decodeFloat(self.B)
-        return self.decode(B + A * self.y / self.z) ** 2
-        # https://github.com/bancorprotocol/carbon-simulator/blob/beta/benchmark/core/trade/impl.py
-        # 'marginalRate' : decodeRate(B + A if y == z else B + A * y / z),
+        # return 0
+        else:
+            return self.decodeRate(self.decodeFloat(int(self.B)) + (self.decodeFloat(int(self.A)) * self.y/self.z))
```
barakman, line 199: And for clarity, you can compute the following before doing the if/else:
- `A = self.decodeFloat(int(self.A))`
- `B = self.decodeFloat(int(self.B))`

fastlane_bot/helpers/poolandtokens.py
```diff
@@ -476,6 +499,66 @@ def decimal_converter(idx):
...
+            no_limit_orders = (strategy_typed_args[0]['is_limit_order'] == False) and (strategy_typed_args[1]['is_limit_order'] == False)
+
+            # evaluate if the price boundaries pa/pb overlap at one end # TODO check logic and remove duplicate logic if necessary
+            prices_overlap = check_overlap(strategy_typed_args[0]['pa'], strategy_typed_args[0]['pb'], strategy_typed_args[1]['pa'], strategy_typed_args[1]['pb'])
+            # if (percent_component_met and no_limit_orders) and not prices_overlap:
+            #     print(percent_component_met, no_limit_orders, prices_overlap)
+            #     print(strategy_typed_args)
+
+            # if the threshold is met and neither is a limit order and prices overlap then likely to be overlapping
+            is_overlapping = percent_component_met and no_limit_orders and prices_overlap
```
barakman, line 528: A strategy with one limit order and one range order can still constitute an overlapping strategy.
In fact, even a strategy with two limit orders can still constitute an overlapping strategy (despite of it being senseless on behalf of its owner to create such strategy).

barakman (changes requested):

fastlane_bot/utils.py
NIXBNT: fixed and cleaned up

fastlane_bot/utils.py
NIXBNT: added

fastlane_bot/helpers/poolandtokens.py
NIXBNT: Indeed but outside of the scope of this work so not required here

barakman (approved):

PR #596, opened 2024-05-02
Remove curve container property

sklbancor:
The `curve_container` property has been deprecated a while ago and is now called `curves`. It shall be removed.


barakman (changes requested):
Empty file `resources/NBTest/BRANCH_REMOVE_CURVE_CONTAINER` added here

PR #598, opened 2024-05-02
Chore/optimizer encapsulated

platonfloria:
Optimizer moved to a separate repo


sklbancor (changes requested):
I would not do that; you change the import statements and this will throw massive conflicts with the changes of the import statements due to restructuring of the repo

sklbancor (approved):

PR #606, opened 2024-05-03
Add sei

platonfloria:


run_blockchain_terraformer.py
```diff
@@ -68,18 +69,20 @@ async def gather():
...
+    "ethereum": 50000,
+    "polygon": 250000,
+    "polygon_zkevm": 500000,
+    "arbitrum_one": 500000,
+    "optimism": 500000,
+    "coinbase_base": 250000,
+    "fantom": 2000,
+    "mantle": 10000000,
+    "linea": 1000000,
+    "sei": 10000,  #TODO untested this could be more
```
barakman, line 85: All but the new one (sei) should remain as is

barakman (changes requested):

run_blockchain_terraformer.py
```diff
@@ -1216,6 +1221,7 @@ def terraform_blockchain(network_name: str):
         file_desc.writelines(list(dict.fromkeys(lines)))
         file_desc.close()
 
+#terraform_blockchain(network_name="sei", save_tokens=True)
```
barakman, line 1224: Should be added below like all the others

barakman (changes requested):

fastlane_bot/config/multicaller.py
```diff
@@ -113,7 +113,7 @@ class MultiCaller(ContextManager):
 
     def __init__(self, contract: web3.contract.Contract,
                  web3: Web3,
-                 block_identifier: Any = 'latest', multicall_address = "0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696"):
+                 block_identifier: Any = 'latest', multicall_address = "0xcA11bde05977b3631167028862bE2a173976CA11"):
```
platonfloria, line 116: What is this address and why was it changed?

platonfloria: Let's replace with MULTICALL_CONTRACT_ADDRESS

resources/NBTest/NBTest_003_Serialization.py
```diff
@@ -205,7 +204,16 @@
 assert raises(CPC.from_carbon, yint=1, y=1, pa=1800, pb=2200, A=100, pair="ETH/USDC", tkny="ETH", fee=0, cid="1", descr="Carbon", isdydx=False)
 assert raises(CPC.from_carbon, yint=1, y=1, pa=1800, pb=2200, B=100, pair="ETH/USDC", tkny="ETH", fee=0, cid="1", descr="Carbon", isdydx=False)
 assert raises(CPC.from_carbon, yint=1, y=1, pa=1800, pb=2200, A=100, B=100, pair="ETH/USDC", tkny="ETH", fee=0, cid="1", descr="Carbon", isdydx=False)
-assert raises(CPC.from_carbon, yint=1, y=1, pb=1800, pa=2200, pair="ETH/USDC", tkny="ETH", fee=0, cid="1", descr="Carbon", isdydx=False)
+#assert raises(CPC.from_carbon, yint=1, y=1, pb=1800, pa=2200, pair="ETH/USDC", tkny="ETH", fee=0, cid="1", descr="Carbon", isdydx=False)
```
platonfloria, line 207: We should avoid committing dead code

fastlane_bot/config/multicaller.py
zavelevsky, line 116: @barakman please make sure it's the right move to replace the default

fastlane_bot/helpers/poolandtokens.py
```diff
@@ -403,10 +403,25 @@ def _carbon_to_cpc(self) -> ConstantProductCurve:
         allow to omit yint (in which case it is set to y, but this does not make
         a difference for the result)
         """
+        def calculate_parameters(y: Decimal, pa: Decimal, pb: Decimal, pm: Decimal, n: Decimal):
```
zavelevsky, line 406: missing tests for this functionality - check_overlap and "cleanup" of overlapping strategies

fastlane_bot/config/multicaller.py
barakman: Verified (the only place where this default value is used is in test 899).
Ideally it should be removed, but since there's a default-value argument which appears before this argument, removing the default value of this argument would require removing the default value of the other argument, or replacing the order in which they are passed to the function.

PR #622, opened 2024-05-06
401 add support for intra and inter carbon deployment arbs

NIXBNT:
- Added specific support for carbon only curves in the multi_pairwise_all (may break tests due to more arbs)
- Expanded the scope of combos (may break tests due to more arbs)
- Added new mode to improve coverage of carbon-centric triangles (new tests tbd)

### DIFF
--- a/fastlane_bot/modes/base_pairwise.py
+++ b/fastlane_bot/modes/base_pairwise.py
@@ -49,10 +49,10 @@ class ArbitrageFinderPairwiseBase(ArbitrageFinderBase):
 
         """
         all_tokens = CCm.tokens()
-        # flashloan_tokens_intersect = all_tokens.intersection(set(flashloan_tokens))
+        flashloan_tokens_intersect = all_tokens.intersection(set(flashloan_tokens))
         combos = [
             (tkn0, tkn1)
-            for tkn0, tkn1 in itertools.product(all_tokens, flashloan_tokens)
+            for tkn0, tkn1 in itertools.product(all_tokens, flashloan_tokens_intersect)
             # tkn1 is always the token being flash loaned
             # note that the pair is tkn0/tkn1, ie tkn1 is the quote token
             if tkn0 != tkn1
--- a/fastlane_bot/modes/base_pairwise.py
+++ b/fastlane_bot/modes/base_pairwise.py
@@ -49,10 +49,10 @@ class ArbitrageFinderPairwiseBase(ArbitrageFinderBase):
 
         """
         all_tokens = CCm.tokens()
-        flashloan_tokens_intersect = all_tokens.intersection(set(flashloan_tokens))
+        # flashloan_tokens_intersect = all_tokens.intersection(set(flashloan_tokens))
         combos = [
             (tkn0, tkn1)
-            for tkn0, tkn1 in itertools.product(all_tokens, flashloan_tokens_intersect)
+            for tkn0, tkn1 in itertools.product(all_tokens, flashloan_tokens)
             # tkn1 is always the token being flash loaned
             # note that the pair is tkn0/tkn1, ie tkn1 is the quote token
             if tkn0 != tkn1
--- a/fastlane_bot/modes/base_triangle.py
+++ b/fastlane_bot/modes/base_triangle.py
@@ -174,6 +174,141 @@ class ArbitrageFinderTriangleBase(ArbitrageFinderBase):
                             combos,
                         )
         return combos
+    
+    def sort_pairs(self, pairs):
+        # Clean up the pairs alphabetically
+        return ["/".join(sorted(pair.split('/'))) for pair in pairs]
+
+    def flatten_nested_items_in_list(self, nested_list):
+        # unpack nested items
+        flattened_list = []
+        for items in nested_list:
+            flat_list = []
+            for item in items:
+                if isinstance(item, list):
+                    flat_list.extend(item)
+                else:
+                    flat_list.append(item)
+            flattened_list.append(flat_list)
+        return flattened_list
+    
+    def get_triangle_groups(self, flt, x_y_pairs):
+        # Get groups of triangles that conform to (flt/x , x/y, y/flt) where x!=y
+        triangle_groups = []
+        for pair in x_y_pairs:
+            x,y = pair.split('/')
+            triangle_groups += [("/".join(sorted([flt,x])), pair, "/".join(sorted([flt,y])))]
+        return triangle_groups
+
+    def get_all_relevant_pairs_info(self, CCm, all_relevant_pairs):
+        # Get pair info for the cohort to allow decision making at the triangle level
+        all_relevant_pairs_info = {}
+        for pair in all_relevant_pairs:            
+            all_relevant_pairs_info[pair] = {}
+            pair_curves = CCm.bypair(pair)
+            carbon_curves = []
+            non_carbon_curves = []
+            for x in pair_curves:
+                if x.params.exchange in self.ConfigObj.CARBON_V1_FORKS:
+                    carbon_curves += [x]
+                else:
+                    non_carbon_curves += [x]
+            all_relevant_pairs_info[pair]['non_carbon_curves'] = non_carbon_curves
+            all_relevant_pairs_info[pair]['carbon_curves'] = carbon_curves
+            all_relevant_pairs_info[pair]['curves'] = non_carbon_curves + [carbon_curves] if len(carbon_curves) >0 else non_carbon_curves  # condense carbon curves into a single list
+            all_relevant_pairs_info[pair]['all_counts'] = len(pair_curves)
+            all_relevant_pairs_info[pair]['carbon_counts'] = len(carbon_curves)
+        return all_relevant_pairs_info
+
+    def get_triangle_groups_stats(self, triangle_groups, all_relevant_pairs_info, arb_mode=None):
+        # Get the stats on the triangle group cohort for decision making
+        triangle_group_stats = {}
+        for i,triangle in enumerate(triangle_groups):
+            triangle_group_stats[i] = {}
+            triangle_group_stats[i]['path'] = []
+            triangle_group_stats[i]['has_path'] = False
+            triangle_group_stats[i]['has_carbon'] = False
+            for pair in triangle:
+                if all_relevant_pairs_info[pair]['all_counts'] > 0:
+                    triangle_group_stats[i]['path'] += [1]
+                else:
+                    triangle_group_stats[i]['path'] += [0]
+                if all_relevant_pairs_info[pair]['carbon_counts'] > 0:
+                    triangle_group_stats[i]['has_carbon'] = True
+            if sum(triangle_group_stats[i]['path']) == 3:
+                triangle_group_stats[i]['has_path'] = True
+
+        # identify valid paths for the given mode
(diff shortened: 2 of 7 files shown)


fastlane_bot/modes/base_triangle.py
```diff
@@ -174,6 +174,141 @@ def get_combos(
...
+            flattened_list.append(flat_list)
+        return flattened_list
+    
+    def get_triangle_groups(self, flt, x_y_pairs):
+        # Get groups of triangles that conform to (flt/x , x/y, y/flt) where x!=y
+        triangle_groups = []
+        for pair in x_y_pairs:
+            x,y = pair.split('/')
+            triangle_groups += [("/".join(sorted([flt,x])), pair, "/".join(sorted([flt,y])))]
+        return triangle_groups
```
barakman, line 201: Why not simply `return self.sort_pairs(x_y_pairs)`?

fastlane_bot/modes/base_triangle.py
```diff
@@ -174,6 +174,141 @@ def get_combos(
...
+            carbon_curves = []
+            non_carbon_curves = []
+            for x in pair_curves:
+                if x.params.exchange in self.ConfigObj.CARBON_V1_FORKS:
+                    carbon_curves += [x]
+                else:
+                    non_carbon_curves += [x]
+            all_relevant_pairs_info[pair]['non_carbon_curves'] = non_carbon_curves
+            all_relevant_pairs_info[pair]['carbon_curves'] = carbon_curves
+            all_relevant_pairs_info[pair]['curves'] = non_carbon_curves + [carbon_curves] if len(carbon_curves) >0 else non_carbon_curves  # condense carbon curves into a single list
```
barakman, line 218: 1. A space is missing in `>0`
2. I'm pretty sure that you should change `[carbon_curves]` to `carbon_curves`
3. It looks like you can get rid of the entire if/else here, and just leave the first part (left of the `if`)

In 3, I'm assuming that you've meant the `+` to be of higher precedence than the `if`.
So you might want to verify this, assuming that you did not intend to get something like:
```
non_carbon_curves + non_carbon_curves
```

barakman (changes requested):

fastlane_bot/modes/base_triangle.py
```diff
@@ -174,6 +174,141 @@ def get_combos(
                             combos,
                         )
         return combos
+    
+    def sort_pairs(self, pairs):
```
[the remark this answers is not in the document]
NIXBNT: updated

fastlane_bot/modes/base_triangle.py
NIXBNT: I think you have misunderstood the function. The triangle_groups now has three pairs in the list

fastlane_bot/modes/base_triangle.py
NIXBNT: 1. Fixed
2. No the carbon curves are condensed to a single list (hence the comment at the end of the line
3. The desired behavior is to populate the dictionary as follows:
     if there are no carbon_curves then simply put in non_carbon_curves
    if there are carbon_curves then condense to a single list ([carbon_curves]) and add it to the non_carbon_curves list
Happy to hear suggestion though I did verify the behavior

fastlane_bot/modes/pairwise_multi_all.py
```diff
@@ -67,6 +67,9 @@ def find_arbitrage(self, candidates: List[Any] = None, ops: Tuple = None, best_p
                 if len(base_direction_two) > 0:
                     curve_combos += [[curve] + base_direction_two for curve in not_carbon_curves]
 
+            if len(carbon_curves) >= 2:
+                curve_combos += [carbon_curves]
```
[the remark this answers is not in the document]
NIXBNT: Fixed

fastlane_bot/modes/base_triangle.py
```diff
@@ -174,6 +174,141 @@ def get_combos(
...
+                else:
+                    triangle_group_stats[i]['path'] += [0]
+                if all_relevant_pairs_info[pair]['carbon_counts'] > 0:
+                    triangle_group_stats[i]['has_carbon'] = True
+            if sum(triangle_group_stats[i]['path']) == 3:
+                triangle_group_stats[i]['has_path'] = True
+
+        # identify valid paths for the given mode
+        valid_carbon_triangles = [triangle_groups[i] for i in triangle_group_stats.keys() if (triangle_group_stats[i]['has_path'] == True) and (triangle_group_stats[i]['has_carbon'] == True)]
+        return valid_carbon_triangles
```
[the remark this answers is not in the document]
NIXBNT: Thanks this is much niced. Updated this and moved all appropriate functions to static methods

fastlane_bot/modes/triangle_multi_complete.py
```diff
@@ -0,0 +1,114 @@
...
+            candidates = []
+
+        combos = self.get_comprehensive_triangles(
+            self.flashloan_tokens, self.CCm, arb_mode=self.arb_mode
+        )
+
+        for src_token, miniverse in combos:
+            # print(miniverse)
+            self.ConfigObj.logger.debug(f"{[x.cid for x in miniverse]}")
+            try:
```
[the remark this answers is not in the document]
NIXBNT: I completely agree. 
I have taken this templated from triangle_multi.py so if we plan to make changes to this try/except loop we should do those changes consistently here and for all modes - and shouldn't update it in this PR.
In any case, I have clean up some print/debug statements

fastlane_bot/modes/triangle_multi_complete.py
```diff
@@ -0,0 +1,114 @@
...
+        pstart = {}
+        for tkn0 in tkn0list:
+            try:
+                price = CCm.bytknx(tkn0).bytkny(tkn1)[0].p
+            except:
+                try:
+                    price = 1/CCm.bytknx(tkn1).bytkny(tkn0)[0].p
+                except Exception as e:
+                    print(str(e))
+                    self.ConfigObj.logger.debug(f"[pstart build] {tkn0} not supported. w {tkn1} {str(e)}")
```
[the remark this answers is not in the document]
NIXBNT: Fixed

fastlane_bot/modes/triangle_multi_complete.py
```diff
@@ -0,0 +1,114 @@
...
+            pstart[tkn0]=price
```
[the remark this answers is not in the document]
NIXBNT: Thanks fixed.

fastlane_bot/modes/base_triangle.py
barakman: I'll have another look then...

barakman (changes requested):
Test 64 currently failing:
```
        assert len(r) >= 25, f"[NBTest64 TestMultiPairwiseAll Mode] Expected at least 25 arbs, found {len(r)}"
        assert multi_carbon_count > 0, f"[NBTest64 TestMultiPairwiseAll Mode] Not finding arbs with multiple Carbon curves."
>       assert carbon_wrong_direction_count == 0, f"[NBTest64 TestMultiPairwiseAll Mode] Expected all Carbon curves to have the same tkn in and tkn out. Mixing is currently not supported."
E       AssertionError: [NBTest64 TestMultiPairwiseAll Mode] Expected all Carbon curves to have the same tkn in and tkn out. Mixing is currently not supported.
E       assert 6 == 0
E         +6
E         -0
```

fastlane_bot/modes/pairwise_multi_all.py
```diff
@@ -67,6 +67,9 @@ def find_arbitrage(self, candidates: List[Any] = None, ops: Tuple = None, best_p
                 if len(base_direction_two) > 0:
                     curve_combos += [[curve] + base_direction_two for curve in not_carbon_curves]
 
+                if len(carbon_curves) >= 2:
+                    curve_combos += [carbon_curves]
```
barakman, line 71: This causes test 064 to fail with:
```
>       assert carbon_wrong_direction_count == 0, f"[NBTest64 TestMultiPairwiseAll Mode] Expected all Carbon curves to have the same tkn in and tkn out. Mixing is currently not supported."
E       AssertionError: [NBTest64 TestMultiPairwiseAll Mode] Expected all Carbon curves to have the same tkn in and tkn out. Mixing is currently not supported.
E       assert 6 == 0
```
We can fix that specific assertion in the test, but I'd like to make sure that this is the right thing to do:
1. Should we expect carbon_wrong_direction_count == 6 here?
2. Does this mean that Mixing is currently not supported is no longer the case?

If not, then there is likely a problem with adding these two lines of code at this specific point.

barakman (changes requested):

fastlane_bot/modes/pairwise_multi_all.py
NIXBNT: Resolved. Can support mixing of carbons, potentially all curves

fastlane_bot/tests/test_064_TestMultiAllMode.py
```diff
@@ -224,7 +224,7 @@ def test_test_expected_output():
     
     assert len(r) >= 25, f"[NBTest64 TestMultiPairwiseAll Mode] Expected at least 25 arbs, found {len(r)}"
     assert multi_carbon_count > 0, f"[NBTest64 TestMultiPairwiseAll Mode] Not finding arbs with multiple Carbon curves."
-    assert carbon_wrong_direction_count == 0, f"[NBTest64 TestMultiPairwiseAll Mode] Expected all Carbon curves to have the same tkn in and tkn out. Mixing is currently not supported."
+    # assert carbon_wrong_direction_count == 0, f"[NBTest64 TestMultiPairwiseAll Mode] Expected all Carbon curves to have the same tkn in and tkn out. Mixing is currently not supported."
```
barakman, line 227: That's not an ideal way to handle it.
Instead of masking out a test, you should make it work.

barakman (changes requested):

fastlane_bot/modes/base_pairwise.py
```diff
@@ -49,10 +49,10 @@ def get_combos(
 
         """
         all_tokens = CCm.tokens()
-        flashloan_tokens_intersect = all_tokens.intersection(set(flashloan_tokens))
+        # flashloan_tokens_intersect = all_tokens.intersection(set(flashloan_tokens))
```
zavelevsky, line 52: this is probably wrong to remove as we want to avoid using a flashloan token we don't really know about

fastlane_bot/tests/test_064_TestMultiAllMode.py
NIXBNT: There is no way to modify this test such that it works.
The underlying principal is that pairs in the "wrong direction" can be supported for intra Carbon, as is the focus of this PR, but also potentially for all pairs. The latter requires more research and, if true, a redesign and optimization of the pairs mode - outside of the scope of this PR.
So TLDR to add support for intra carbon we need to pass "wrong direction" carbon pairs and so this breaks the test

fastlane_bot/modes/base_pairwise.py
NIXBNT: Yes agreed will revert this change.
I suspect that the confusion came from the fact that if flashloan tokens are not checksummed they get dropped at this stage, will open a separate issue for that

PR #638, opened 2024-05-09
checksum flashloan and target tokens

NIXBNT:
Improve the user experience by checksumming input tokens so they arent dropped later

### DIFF
--- a/fastlane_bot/events/utils.py
+++ b/fastlane_bot/events/utils.py
@@ -233,7 +233,6 @@ def get_tkn_symbols(flashloan_tokens, tokens: pd.DataFrame) -> List:
 
     returns: list
     """
-    flashloan_tokens = flashloan_tokens.split(",")
     flashloan_tkn_symbols = []
     for tkn in flashloan_tokens:
         flashloan_tkn_symbols.append(get_tkn_symbol(tkn_address=tkn, tokens=tokens))
@@ -505,6 +504,7 @@ def handle_target_tokens(
             target_tokens = flashloan_tokens
         else:
             target_tokens = target_tokens.split(",")
+            target_tokens = [cfg.w3.to_checksum_address(x) for x in target_tokens]
             # target_tokens = [
             #     QueryInterface.cleanup_token_key(token) for token in target_tokens
             # ]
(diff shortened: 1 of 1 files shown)


fastlane_bot/events/utils.py
```diff
@@ -233,7 +233,6 @@ def get_tkn_symbols(flashloan_tokens, tokens: pd.DataFrame) -> List:
 
     returns: list
     """
-    flashloan_tokens = flashloan_tokens.split(",")
```
NIXBNT, line 236: drop this because its done before this function

fastlane_bot/events/utils.py
```diff
@@ -542,8 +542,9 @@ def handle_flashloan_tokens(
     List[str]
         A list of flashloan tokens to fetch data for.
     """
-    flt_symbols = get_tkn_symbols(flashloan_tokens=flashloan_tokens, tokens=tokens)
     flashloan_tokens = flashloan_tokens.split(",")
+    flashloan_tokens = [cfg.w3.to_checksum_address(x) for x in flashloan_tokens]
+    flt_symbols = get_tkn_symbols(flashloan_tokens=flashloan_tokens, tokens=tokens)
```
NIXBNT, line 547: Needed to change the order here else the token symbols are fetched on the checksummed addresses

fastlane_bot/events/utils.py
```diff
@@ -638,8 +639,8 @@ def get_config(
     cfg.LIMIT_BANCOR3_FLASHLOAN_TOKENS = limit_bancor3_flashloan_tokens
     cfg.DEFAULT_MIN_PROFIT_GAS_TOKEN = Decimal(default_min_profit_gas_token)
     cfg.GAS_TKN_IN_FLASHLOAN_TOKENS = (
-        cfg.NATIVE_GAS_TOKEN_ADDRESS in flashloan_tokens
-        or cfg.WRAPPED_GAS_TOKEN_ADDRESS in flashloan_tokens
+        cfg.NATIVE_GAS_TOKEN_ADDRESS.lower() in flashloan_tokens.lower()
+        or cfg.WRAPPED_GAS_TOKEN_ADDRESS.lower() in flashloan_tokens.lower()
     )
```
NIXBNT, line 644: Since the w3 module is required for checksum, and the cfg is required to be initialized before using w3, its not as clean to checksum the flashloan_tokens AND target_tokens at this point.
i.e. if we want to checksum flashloan tokens here it can go after w3 intialized but then we also need to pass the target tokens - when we have specific functions called handle_flashloan_tokens() etc.
So if the native or wrapped gas token are not  checksummed we can check for their presence with lower() instead without modifying

PR #640, opened 2024-05-09
Refactor PR 622

barakman:


barakman:
Partially reposted as part of PR #648 (except for the simplification around the handling of `non_carbon_curves` and `carbon_curves`, which requires some more careful thought).

PR #647, opened 2024-05-12
Fix the pool-finder-period handling

barakman:


fastlane_bot/events/exchanges/base.py
```diff
@@ -124,13 +123,6 @@ async def get_fee(address: str, contract: AsyncContract) -> float:
         """
         pass
 
-    @abstractmethod
```
Lesigh-3100, line 127: This function needs to be renamed & modified instead of removed because the `get_pool` function that it's now using is something completely different.

barakman: I don't understand, the function `get_pool_with_multicall` was renamed to `get_pool` in every class in which it was actually useful, and removed in all other classes. Why does it need to be renamed here instead of removed?

fastlane_bot/pool_finder.py
```diff
@@ -108,18 +85,17 @@ def _find_pools(self, unsupported_pairs: List[Tuple]) -> Dict[str, List[str]]:
                 mc = MultiCaller(contract=exchange.factory_contract, web3=self._web3, multicall_address=self._multicall_address)
                 for pair in pair_chunk:
                     if exchange.base_exchange_name in [UNISWAP_V2_NAME, SOLIDLY_V2_NAME]:
-                        exchange.get_pool_with_multicall(mc, pair[0], pair[1])
+                        mc.add_call(exchange.get_pool_func_call(pair[0], pair[1]))
```
Lesigh-3100, line 88: Unfortunately the way `mc.add_call` and `get_pool_func_call` was changed doesn't work. 

`mc.add_call` is expecting the following format for args: 
`mc.add_call(FUNCTION_FROM_CONTRACT, args)`

For example:
`mc.add_call(factory_contract.functions.getPair, tkn0, tkn1, False)`

So to remove the Multicaller dependency, we would need to change the Exchange `get_pool_func_call` to return the function with no args and modify the main loop here - likely also splitting apart Uni V2 & Solidy because they have different inputs. 

If what I wrote doesn't make enough sense please LMK!

barakman: That's how it works in the new MC (PR #641); I'll update this branch accordingly.

Issue #650, opened 2024-05-13
Refactor Triangle Mode


barakman:
The triangle-mode feature is currently implemented in 3 separate modules, consisting of partially-duplicated pieces of code and fully-duplicated pieces of code.

It should be refactored as follows:

---

Class `ArbitrageFinderTriangleBase` should implement function `find_arbitrage`, which will call:
- Function `assert_supported`
- Function `get_combos`

The rest of function `find_arbitrage` should be identical to how it is implemented in each one of the derived classes.

---

Class `ArbitrageFinderTriangleMulti` should implement:
- Function `assert_supported` to raise an exception if `self.base_exchange != "carbon_v1"`
- Function `get_combos` as this function is currently implemented in the base class

All the functionality related to the latter should be moved out of the base class and into this derived class.

---

Class `ArbitrageFinderTriangleMultiComplete` should implement:
- Function `assert_supported` to pass
- Function `get_combos` as function `get_comprehensive_triangles` is currently implemented in the base class

All the functionality related to the latter should be moved out of the base class and into this derived class.

PR #651, opened 2024-05-13
Refactor triangle-mode

barakman:
Handle issue #650


fastlane_bot/modes/base_triangle.py
```diff
@@ -9,301 +9,83 @@
...
-        if miniverses:
-            combos += list(zip([flt] * len(miniverses), miniverses))
-        return combos
-
-    def get_combos(
-        self, flashloan_tokens: List[str], CCm: Any, arb_mode: str
-    ) -> Tuple[List[str], List[Any]]:
-        """
-        Get combos for triangular arbitrage
+    def assert_supported(self):
```
sklbancor, line 25: what is this meant to do? doc string would he helpful

fastlane_bot/modes/base_triangle.py
```diff
@@ -9,301 +9,83 @@
...
-        return flt_triangle_analysis_set
-
-    def get_comprehensive_triangles(
-        self, flashloan_tokens: List[str], CCm: Any
-    ) -> Tuple[List[str], List[Any]]:
-        """
-        Get comprehensive combos for triangular arbitrage
+    @abc.abstractmethod
+    def get_combos(self, flashloan_tokens: List[str], CCm: Any, arb_mode: str) -> Tuple[List[str], List[Any]]:
+        ...
```
sklbancor, line 30: doc string? generally in abstract methods it is good to not only define the signature but also the semantics.

fastlane_bot/modes/base_triangle.py
```diff
@@ -9,301 +9,83 @@
...
 
-        Parameters
-        ----------
-        flashloan_tokens : list
-            List of flashloan tokens
-        CCm : object
-            CCm object
+    def find_arbitrage(self, candidates: List[Any] = None, ops: Tuple = None, best_profit: float = 0, profit_src: float = 0) -> Union[List, Tuple]:
```
sklbancor, line 32: doc string

fastlane_bot/modes/base_triangle.py
```diff
@@ -9,301 +9,83 @@
...
+                CC_cc = CPCContainer(miniverse)
+                O = MargPOptimizer(CC_cc)
+                pstart = self.build_pstart(CC_cc, CC_cc.tokens(), src_token)
+                r = O.optimize(src_token, params=dict(pstart=pstart))
+                trade_instructions_dic = r.trade_instructions(O.TIF_DICTS)
+                if trade_instructions_dic is None or len(trade_instructions_dic) < 3:
+                    # Failed to converge
+                    continue
+                trade_instructions_df = r.trade_instructions(O.TIF_DFAGGR)
+                trade_instructions = r.trade_instructions()
```
sklbancor, line 51: why do you get the trade instructions in all those different formats? they are equivalent and you should choose one

sklbancor (changes requested):

fastlane_bot/modes/triangle_multi.py
```diff
@@ -22,65 +20,141 @@ class ArbitrageFinderTriangleMulti(ArbitrageFinderTriangleBase):
...
+        if arb_mode in ["b3_two_hop"]:
+            combos = [
+                (tkn0, tkn1)
+                for tkn0, tkn1 in product(flashloan_tokens, flashloan_tokens)
+                # note that the pair is tkn0/tkn1, ie tkn1 is the quote token
+                if tkn0 != tkn1
+            ]
+        else:
+            all_base_exchange_curves = CCm.byparams(exchange=self.base_exchange).curves
+            for flt in flashloan_tokens:  # may wish to run this for one flt at a time
```
sklbancor, line 55: I don't understand the comment?

fastlane_bot/modes/triangle_multi.py
```diff
@@ -22,65 +20,141 @@ class ArbitrageFinderTriangleMulti(ArbitrageFinderTriangleBase):
...
+        combos = []
+        if arb_mode in ["b3_two_hop"]:
+            combos = [
+                (tkn0, tkn1)
+                for tkn0, tkn1 in product(flashloan_tokens, flashloan_tokens)
+                # note that the pair is tkn0/tkn1, ie tkn1 is the quote token
+                if tkn0 != tkn1
+            ]
+        else:
+            all_base_exchange_curves = CCm.byparams(exchange=self.base_exchange).curves
```
sklbancor, line 54: comments on what you are doing and why you are doing it would be helpful

fastlane_bot/modes/triangle_multi.py
```diff
@@ -22,65 +20,141 @@ class ArbitrageFinderTriangleMulti(ArbitrageFinderTriangleBase):
...
+                        .byparams(exchange=self.base_exchange)
+                        .curves
+                    )
+                    if len(base_exchange_curves) == 0:
+                        continue
+
+                    base_direction_pair = base_exchange_curves[0].pair
+                    base_direction_one = [curve for curve in base_exchange_curves if curve.pair == base_direction_pair]
+                    base_direction_two = [curve for curve in base_exchange_curves if curve.pair != base_direction_pair]
+                    assert len(base_exchange_curves) == len(base_direction_one) + len(base_direction_two)
```
sklbancor, line 73: cant really fail, no?

fastlane_bot/modes/triangle_multi.py
```diff
@@ -22,65 +20,141 @@ class ArbitrageFinderTriangleMulti(ArbitrageFinderTriangleBase):
...
+                    y_match_curves = CCm.bypairs(
```
sklbancor, line 74: y takes tknx and vice versa versa. correct?

fastlane_bot/modes/base_triangle.py
barakman: Outside the scope of this PR (same code used prior to it)

fastlane_bot/modes/triangle_multi.py
barakman: Outside the scope of this PR (same code used prior to it)

fastlane_bot/modes/triangle_multi.py
barakman: Outside the scope of this PR (same code used prior to it)

fastlane_bot/modes/base_triangle.py
sklbancor: but are they all used below?

fastlane_bot/modes/triangle_multi.py
barakman: Outside the scope of this PR (same code used prior to it), but good point regardless, so will remove

fastlane_bot/modes/base_triangle.py
barakman: Abstract method removed

fastlane_bot/modes/base_triangle.py
barakman: Outside the scope of this PR (same code used prior to it)

barakman: That's right. I'll add doc-strings as soon as i complete this refactoring (which currently just keeps growing).

fastlane_bot/modes/base_triangle.py
barakman: Annoying indeed.
I was initially aiming for the least amount of changes required, but as this task continues to expand, it is becoming more and more reasonable that we may as well clean up this sort of stuff (which seems to exists in each and everyone of the modes, BTW).

fastlane_bot/modes/triangle_multi.py
```diff
@@ -8,79 +8,72 @@
...
+        all_carbon_curves = self.CCm.byparams(exchange="carbon_v1").curves
+
+        for flt in self.flashloan_tokens:
+            non_flt_carbon_curves = [curve for curve in all_carbon_curves if flt not in curve.pair]
+            for non_flt_carbon_curve in non_flt_carbon_curves:
+                tkny = non_flt_carbon_curve.tkny
+                tknx = non_flt_carbon_curve.tknx
+
+                carbon_curves = self.CCm.bypairs(f"{tknx}/{tkny}").byparams(exchange="carbon_v1").curves
+                if len(carbon_curves) == 0:
```
Lesigh-3100, line 28: The check for carbon curves here should be removed - there is no need to fail this mode if no Carbon Curves exist.

fastlane_bot/modes/triangle_bancor_v3_two_hop.py
```diff
@@ -8,362 +8,54 @@
...
-            # Get the candidate ids
-            cids = [ti["cid"] for ti in trade_instructions_dic]
-
-            # Calculate the profit
-            profit = self.calculate_profit(src_token, profit_src, self.CCm, cids)
 
-            if str(profit) == "nan":
-                self.ConfigObj.logger.debug("profit is nan, skipping")
+            external_curves = [curve for curve in all_curves if curve.params.get("exchange") not in self.ConfigObj.CARBON_V1_FORKS]
+            if len(external_curves) == 0:
```
Lesigh-3100, line 36: This check and the check for carbon_curves above should be combined to:
```
if len(external_curves) == 0 and len(carbon_curves) == 0:
    continue
```

fastlane_bot/modes/triangle_bancor_v3_two_hop.py
```diff
@@ -8,362 +8,54 @@
...
+                miniverses += [bancor_v3_curve_0 + bancor_v3_curve_1 + base_dir_one]
 
-                    if len(base_direction_one) > 0:
-                        miniverses += [bancor_v3_curve_0 + bancor_v3_curve_1 + base_direction_one]
+            if len(base_dir_two) > 0:
+                miniverses += [bancor_v3_curve_0 + bancor_v3_curve_1 + base_dir_two]
 
-                    if len(base_direction_two) > 0:
-                        miniverses += [bancor_v3_curve_0 + bancor_v3_curve_1 + base_direction_two]
+            miniverses += [bancor_v3_curve_0 + bancor_v3_curve_1 + carbon_curves]
```
Lesigh-3100, line 57: This should be removed as the carbon curves are being added separately above.

barakman: Fixed

fastlane_bot/modes/triangle_bancor_v3_two_hop.py
barakman: If you meant `or`, then the second part of this check is already applied a few lines above.
If you meant `and` then please clarify the reason.

Lesigh-3100: We only want to `continue` if there are no curves to pair with the bancor v3 curves. 

If there are external curves or carbon curves, we want to complete the process.

barakman: Revised, can you please take a look?

fastlane_bot/bot.py
```diff
@@ -264,46 +243,27 @@ def _get_deadline(self, block_number) -> int:
...
-        finder = arb_finder(
-            flashloan_tokens=flashloan_tokens,
-            CCm=CCm,
-            mode="bothin",
-            result=random_mode,
-            ConfigObj=self.ConfigObj,
-        )
-        return {"finder": finder, "r": finder.find_arbitrage()}
+    def get_arb_finder(self, arb_mode: str, flashloan_tokens, CCm) -> ArbitrageFinderBase:
+        return {
```
sklbancor, line 251: not a biggie but I personally would define the dict in a class constant that can be inspected when looking at the class and then reference it here rather than defining it only inline.

This also has the advantage that you can include it in tests, eg you can check that the dict keys of available methods are what you expected they are [removing a key there is probably a semver major version event]

fastlane_bot/bot.py
```diff
@@ -264,46 +243,27 @@ def _get_deadline(self, block_number) -> int:
...
         self,
-        flashloan_tokens: List[str],
-        CCm: CPCContainer,
         *,
-        arb_mode: str,
-        randomizer: int,
-        data_validator=False,
+        flashloan_tokens: List[str] = None,
+        CCm: CPCContainer = None,
+        arb_mode: str = None,
```
sklbancor, line 264: generally I would probably use enums here

fastlane_bot/bot.py
```diff
@@ -312,55 +272,36 @@ def _run(
         Parameters
         ----------
         flashloan_tokens: List[str]
-            The tokens to flashloan.
+            The flashloan tokens
         CCm: CPCContainer
-            The container.
+            The complete market data container
```
sklbancor, line 277: I am not sure of the meaning of "complete" here. Is it the entire market at this stage, or only the curves for the optimization?

fastlane_bot/bot.py
```diff
@@ -312,55 +272,36 @@ def _run(
...
-            elif self.validate_pool_data(arb_opp=r):
-                self.ConfigObj.logger.debug(
-                    "[bot._run] All data checks passed! Pools in sync!"
-                )
-            else:
-                self.ConfigObj.logger.warning(
-                    "[bot._run] Data validation failed. Updating pools and restarting."
-                )
-                return
+        arb_opp = rand_item(list_of_items=arb_opps, num_of_items=randomizer)
```
sklbancor, line 302: this is consistent with what we are doing now, but arguably randomisation should happen at the node level because they have the complete picture; we should just return everything we find

PR #657, opened 2024-05-20
handle curve direction triangle complete

NIXBNT:
Carbon curves should be collected by their direction (i.e. the token that they are selling) to prevent downstream failures building the route

### DIFF
--- a/fastlane_bot/modes/triangle_multi_complete.py
+++ b/fastlane_bot/modes/triangle_multi_complete.py
@@ -85,12 +85,22 @@ def get_all_relevant_pairs_info(CCm, all_relevant_pairs, carbon_v1_forks):
         pair_curves = CCm.bypair(pair)
         carbon_curves = []
         non_carbon_curves = []
+        base_direction_one = []
+        base_direction_two = []
+        base_direction_pair = pair_curves[0].pair
         for x in pair_curves:
             if x.params.exchange in carbon_v1_forks:
                 carbon_curves += [x]
+                if x.pair == base_direction_pair:
+                    base_direction_one += [x]
+                else:
+                    base_direction_two += [x]
             else:
                 non_carbon_curves += [x]
-        all_relevant_pairs_info[pair]['curves'] = non_carbon_curves + [carbon_curves] if len(carbon_curves) > 0 else non_carbon_curves  # condense carbon curves into a single list
+        all_relevant_pairs_info[pair]['curves'] = non_carbon_curves
+        # for each direction, condense carbon curves into a single list and add to the non-carbon curves
+        all_relevant_pairs_info[pair]['curves'] += [base_direction_one] if len(base_direction_one) > 0 else None 
+        all_relevant_pairs_info[pair]['curves'] += [base_direction_two] if len(base_direction_two) > 0 else None     
         all_relevant_pairs_info[pair]['all_counts'] = len(pair_curves)
         all_relevant_pairs_info[pair]['carbon_counts'] = len(carbon_curves)
     return all_relevant_pairs_info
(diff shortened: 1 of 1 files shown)


fastlane_bot/modes/triangle_multi_complete.py
```diff
@@ -85,12 +85,22 @@ def get_all_relevant_pairs_info(CCm, all_relevant_pairs, carbon_v1_forks):
...
+                    base_direction_one += [x]
+                else:
+                    base_direction_two += [x]
             else:
                 non_carbon_curves += [x]
-        all_relevant_pairs_info[pair]['curves'] = non_carbon_curves + [carbon_curves] if len(carbon_curves) > 0 else non_carbon_curves  # condense carbon curves into a single list
+        all_relevant_pairs_info[pair]['curves'] = non_carbon_curves
+        # for each direction, condense carbon curves into a single list and add to the non-carbon curves
+        all_relevant_pairs_info[pair]['curves'] += [base_direction_one] if len(base_direction_one) > 0 else None 
+        all_relevant_pairs_info[pair]['curves'] += [base_direction_two] if len(base_direction_two) > 0 else None
```
barakman, line 103: Why are you adding `None` to `all_relevant_pairs_info[pair]['curves']`?

barakman (changes requested):

fastlane_bot/modes/triangle_multi_complete.py
NIXBNT: udpated. thanks

fastlane_bot/modes/triangle_multi_complete.py
```diff
@@ -85,12 +85,24 @@ def get_all_relevant_pairs_info(CCm, all_relevant_pairs, carbon_v1_forks):
...
+                    base_direction_two += [x]
             else:
                 non_carbon_curves += [x]
-        all_relevant_pairs_info[pair]['curves'] = non_carbon_curves + [carbon_curves] if len(carbon_curves) > 0 else non_carbon_curves  # condense carbon curves into a single list
+        all_relevant_pairs_info[pair]['curves'] = non_carbon_curves
+        # for each direction, condense carbon curves into a single list and add to the non-carbon curves
+        if len(base_direction_one) > 0:
+            all_relevant_pairs_info[pair]['curves'].append(base_direction_one)
+        if len(base_direction_two) > 0:
+            all_relevant_pairs_info[pair]['curves'].append(base_direction_two)
```
barakman, line 105: Are you sure that you want to append the array and not concat it?
Previously, with `+=`, you were adding **its elements**.
Now you are adding it as a single nested-array element.

barakman (changes requested):

fastlane_bot/modes/triangle_multi_complete.py
```diff
@@ -85,12 +85,24 @@ def get_all_relevant_pairs_info(CCm, all_relevant_pairs, carbon_v1_forks):
         pair_curves = CCm.bypair(pair)
         carbon_curves = []
         non_carbon_curves = []
+        base_direction_one = []
+        base_direction_two = []
+        base_direction_pair = pair_curves[0].pair
         for x in pair_curves:
             if x.params.exchange in carbon_v1_forks:
                 carbon_curves += [x]
```
barakman, line 93: What is this used for?

barakman (changes requested):

fastlane_bot/modes/triangle_multi_complete.py
barakman: Well, it was `+= [base_direction_one]` and `+= [base_direction_two]`, so perhaps no issue here.

NIXBNT: The desired result is a nested list.
Previously, while I did use `+=`, I did so with additional brackets to ensure it was nested

fastlane_bot/modes/triangle_multi_complete.py
NIXBNT: The base direction is used to sort carbon orders into the groups of which token (in the pair) is being sold. This way all the orders selling the same asset are grouped together.
This downstream processes are dependent on this.

barakman: I was referring to `carbon_curves`. Is it still used anywhere?

barakman (approved):

PR #658, opened 2024-05-20
feat: get logs batching

platonfloria:


fastlane_bot/events/event_gatherer.py
```diff
@@ -48,16 +51,46 @@ def get_all_events(self, from_block: int, to_block: int):
...
+        if from_block <= to_block:
+            try:
+                return await self._w3.eth.get_logs(filter_params={
+                    "fromBlock": from_block,
+                    "toBlock": to_block,
+                    "topics": topics
+                })
+            except Exception as e:
+                assert "eth_getLogs" in str(e), str(e)
+                mid_block = (from_block + to_block) // 2
```
zavelevsky, line 90: Anything here stopping the code from halving mid block forever in case of some error?

zavelevsky (changes requested):

fastlane_bot/config/constants.py
```diff
@@ -34,3 +34,16 @@
...
+    "ethereum": 0,
+    "polygon": 0,
+    "polygon_zkevm": 0,
+    "arbitrum_one": 0,
+    "optimism": 0,
+    "coinbase_base": 0,
+    "fantom": 5000,
+    "mantle": 0,
+    "linea": 0,
+    "sei": 0,
```
barakman, line 48: Depending on the RPC URL planned for use, this limit may need to change:
- For `https://evm-rpc.arctic-1.seinetwork.io/`, use 0
- For `https://rpc.mantle.xyz`, use 10000

fastlane_bot/events/event_gatherer.py
barakman: No. In order to fix this (also in the terraformer script if still applicable), you should place the part which follows the assertion under `if from_block < to_block`.

barakman (changes requested):

fastlane_bot/config/constants.py
platonfloria: This will require more granular configuration. Something like having a separate config parameters for each RPC, which we should strive for anyway. Let's keep at 0 for now, as that will work for both RPCs. But let's keep in mind that we might need to change it when the mainnet is launched.

fastlane_bot/events/event_gatherer.py
platonfloria: Done

zavelevsky (approved):

fastlane_bot/events/event_gatherer.py
```diff
@@ -48,16 +51,47 @@ def get_all_events(self, from_block: int, to_block: int):
...
+            except Exception as e:
+                assert "eth_getLogs" in str(e), str(e)
+                if from_block < to_block:
+                    mid_block = (from_block + to_block) // 2
+                    log_lists = await asyncio.gather(
+                        self._get_logs_recursive(from_block, mid_block, topics),
+                        self._get_logs_recursive(mid_block + 1, to_block, topics)
+                    )
+                    return [log for log_list in log_lists for log in log_list]
+        return []
```
barakman, line 97: We probably need to raise an error here, because we otherwise may return with missing results but nothing to indicate that.

run_blockchain_terraformer.py
```diff
@@ -680,10 +680,11 @@ def get_events_recursive(get_logs: any, start_block: int, end_block: int) -> lis
...
-            mid_block = (start_block + end_block) // 2
-            event_list_1 = get_events_recursive(get_logs, start_block, mid_block)
-            event_list_2 = get_events_recursive(get_logs, mid_block + 1, end_block)
-            return event_list_1 + event_list_2
+            if start_block < end_block:
+                mid_block = (start_block + end_block) // 2
+                event_list_1 = get_events_recursive(get_logs, start_block, mid_block)
+                event_list_2 = get_events_recursive(get_logs, mid_block + 1, end_block)
+                return event_list_1 + event_list_2
     return []
```
barakman, line 688: Same

barakman (changes requested):

fastlane_bot/events/event_gatherer.py
platonfloria: Let's print exception instead of raising it, so that the process is not interrupted. Done so

run_blockchain_terraformer.py
platonfloria: Ditto

fastlane_bot/events/event_gatherer.py
```diff
@@ -48,16 +52,49 @@ def get_all_events(self, from_block: int, to_block: int):
...
+                assert "eth_getLogs" in str(e), str(e)
+                if from_block < to_block:
+                    mid_block = (from_block + to_block) // 2
+                    log_lists = await asyncio.gather(
+                        self._get_logs_recursive(from_block, mid_block, topics),
+                        self._get_logs_recursive(mid_block + 1, to_block, topics)
+                    )
+                    return [log for log_list in log_lists for log in log_list]
+                else:
+                    traceback.print_exc(e)
```
barakman, line 99: 1. You should use `self.cfg.logger.error` here, or alternatively - raise an exception here, which will be caught at the top-level (the main loop), where that traceback will be logged exactly as you are trying to do here (and even better - it will be logged also in the log file, which you are **not** achieving here).
2. You should just put it at the very end (instead of returning an empty list), so we'd also catch any erroneous scenario (i.e., cases where the initial input to this function yields `from_block > to_block`).

And again, following bullet 1 above, you should raise an exception here rather than inventing some new approach, because the entire system is designated to work this way.

You shouldn't worry too much about any such exception (indicating missing events in the output) taking place very often, because it can really be the result of only two things:
- Network (dis)connection
- Illegal input (`from_block > to_block`)

For each one of these cases, it is completely reasonable to throw the execution of the program back to the beginning of the next iteration. In fact, it is probably even **desirable** to do so.

The only other option for such a scenario is when function `get_logs` fails to retrieve a single block (`from_block == to_block`), which is very unlikely to happen, and it probably justifies throwing the execution of the program back to the beginning of the next iteration just as equally as each one of the other two scenarios.

barakman (changes requested):

fastlane_bot/events/event_gatherer.py
platonfloria: Resolved

fastlane_bot/events/event_gatherer.py
```diff
@@ -30,6 +32,7 @@ def __init__(
             manager: The Manager object
             w3: The connected AsyncWeb3 object.
         """
+        self._blockchain = blockchain
```
barakman, line 35: Can obtain this from `self.cfg.network.NETWORK` or something similar, rather than creating a duplicated attribute.

barakman (changes requested):

fastlane_bot/events/event_gatherer.py
barakman: Or maybe `self.cfg.NETWORK`

platonfloria: EventGatherer has no access to the config object

barakman: OK, then it is probably better to add the Config object rather than the blockchain ID, since it ensures that the same blockchain IDs are used across the code (i.e., it helps preventing mistakes, so one just needs to pass the same Config object which is used everywhere else in the system).

barakman (approved):

PR #659, opened 2024-05-20
Telos Updated

NIXBNT:
`add_telos` branch rebased onto the `sei-release-candidate`
A few things changed that need checking and considering:
- solidly_v2 fork archly_v2 (the solidly handling changed a bit since the initial branch)
- Added support for legacy transaction type with self.cfg.SUPPORTS_EIP1559 (Telos does not support EIP1559)

Terraformer works with eth_logs.
Test transaction successful using legacy transaction


fastlane_bot/helpers/txhelpers.py
```diff
@@ -110,7 +110,10 @@ def validate_and_submit_transaction(
 
         raw_tx = self._sign_transaction(tx)
 
-        gas_cost_wei = tx["gas"] * tx["maxFeePerGas"]
+        if self.cfg.SUPPORTS_EIP1559:
+            gas_cost_wei = tx["gas"] * tx["maxFeePerGas"]
+        else:
+            gas_cost_wei = tx["gas"]
```
barakman, line 116: Need to multiply by `tx["gasPrice"]`

fastlane_bot/helpers/txhelpers.py
```diff
@@ -150,22 +153,32 @@ def check_and_approve_tokens(self, tokens: List):
...
+                "value": value,
+                "chainId": self.chain_id,
+                "from": self.wallet_address,
+                "to": contract.address,
+                "data": contract.encode_abi(fn_name=fn_name, args=args),
+                "nonce": self.cfg.w3.eth.get_transaction_count(self.wallet_address)
+            }
+        else:
+            return {
+                "value": value,
```
barakman, line 175: Missing `type` (0 or 1, need to find out with Telos which one is supported)

barakman (changes requested):

fastlane_bot/helpers/txhelpers.py
NIXBNT: updated

fastlane_bot/helpers/txhelpers.py
NIXBNT: Actually not.
When passing type as 1: ValueError: {'message': 'Invalid Transaction: Calling from_big_endian with oversized array', 'code': 3}
When passing type as 0: TypeError: Unknown Transaction type: 0
When passing nothing as I have it, the arb executes immediately

PR #664, opened 2024-05-22
feat: add python12 support by dropping python8 support

platonfloria:


pyproject.toml
```diff
@@ -18,7 +18,6 @@ python-dotenv = "^0.16.0"
 joblib = "^1.2.0"
 pandas = "^1.5.2"
 alchemy-sdk = "^0.1.1"
```
barakman, line 20: We might be able to get rid of this one as well (need to check)

barakman (approved):

PR #666, opened 2024-05-22
optimize start block

NIXBNT:


fastlane_bot/events/utils.py
```diff
@@ -1293,7 +1293,7 @@ def get_start_block(
         elif mgr.tenderly_fork_id:
             return safe_int(max(block["last_updated_block"] for block in mgr.pool_data)) - reorg_delay, mgr.w3_tenderly.eth.block_number
         else:
-            return safe_int(max(block["last_updated_block"] for block in mgr.pool_data)) - reorg_delay, None
+            return max(safe_int(max(block["last_updated_block"] for block in mgr.pool_data)) , last_block) - reorg_delay, None
```
barakman, line 1296: Isn't it that **by definition** `last_block  >= last_updated_block` for all for blocks in `mgr.pool_data`?
In which case, your code is equivalent to simply returning `last_block`.

From looking into how this function is used, it seems that the returned value is then passed on to function `get_latest_events`, which means that if the value of `last_block` is passed to it, then most likely no events will be fetched (or perhaps only the events in the very last block).

So I suspect this this fix achieves undesired behavior (perhaps it reduces the time spent on fetching blocks, but on the same time, it may be skipping many required events).

barakman (changes requested):

PR #667, opened 2024-05-22
fix: handle timeout exception correctly

platonfloria:


fastlane_bot/events/event_gatherer.py
```diff
@@ -86,7 +86,6 @@ async def _get_logs_recursive(self, from_block: int, to_block: int, topics: List
                     "topics": topics
                 })
             except Exception as e:
-                assert "eth_getLogs" in str(e), str(e)
```
barakman, line 89: Please log any error-message which **doesn't** contain the `"eth_getLogs"` part in it.

This error is well expected as part of this binary-search, but any other error is something that we would want to be aware of.

For example, network disconnection, or even some silly coding mistake inside the `try` part.

Otherwise, we're essentially masking out potential bugs and issues.

run_blockchain_terraformer.py
```diff
@@ -679,7 +679,6 @@ def get_events_recursive(get_logs: any, start_block: int, end_block: int) -> lis
         try:
             return get_logs(fromBlock=start_block, toBlock=end_block)
         except Exception as e:
-            assert "eth_getLogs" in str(e), str(e)
```
barakman, line 682: Same as above

barakman (changes requested):

zavelevsky (approved):

main.py
```diff
@@ -307,7 +307,7 @@ def run(mgr, args, tenderly_uri=None) -> None:
     handle_static_pools_update(mgr)
 
     event_gatherer = EventGatherer(
-        blockchain=mgr.cfg.network.NETWORK,
+        config=mgr.cfg,
         w3=mgr.w3_async,
```
barakman, line 311: Can obtain `w3` internally, via `config.w3` (instead of passing it as an additional input.

barakman (changes requested):

barakman (approved):

PR #673, opened 2024-05-23
Optimizer hotfix

sklbancor:
could fix the excessive logging problem; can not really cause any havoc (famous last words)

### DIFF
--- a/fastlane_bot/tools/optimizer/margpoptimizer.py
+++ b/fastlane_bot/tools/optimizer/margpoptimizer.py
@@ -373,6 +373,13 @@ class MargPOptimizer(CPCArbOptimizer):
                 p = np.exp(plog10 * np.log(10))
                 criterium = np.linalg.norm(dplog10)
                 
+                # this fix could cut down on some of the logging noise by raising a ConvergenceError early
+                if max(p) > 1e50:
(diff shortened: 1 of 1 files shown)


fastlane_bot/tools/optimizer/margpoptimizer.py
```diff
@@ -373,6 +373,13 @@ def dtknfromp_f(p, *, islog10=True, asdct=False, quiet=False):
                 p = np.exp(plog10 * np.log(10))
                 criterium = np.linalg.norm(dplog10)
                 
+                # this fix could cut down on some of the logging noise by raising a ConvergenceError early
+                if max(p) > 1e50:
+                    raise self.ConvergenceError(f"price vector component exceeds maximum price [p={p}]")
+                if min(p) < 1e-50:
+                    raise self.ConvergenceError(f"price vector component below minimum price [p={p}]")
```
barakman, line 380: The 'overflow' warnings are emitted right before you raise this exception, at:
```
p = np.exp(plog10 * np.log(10))
```
Consider raising it prior to that line, by applying a similar check on `plog10` (since `p` is not yet available).

barakman (changes requested):

PR #676, opened 2024-05-27
Triangle mode refactor with pstart

NIXBNT:
New pstart handling (in triangle_compete mode)
Bugfix token handling for NATIVE/WRAPPED
Log clean up
Remove duplicate iterations for univ2

### DIFF
--- a/fastlane_bot/modes/base_triangle.py
+++ b/fastlane_bot/modes/base_triangle.py
@@ -8,6 +8,7 @@ Defines the base class for triangular arbitrage finder modes
 All rights reserved.
 Licensed under MIT.
 """
+import random
 from typing import Any, List, Dict
 from fastlane_bot.tools.cpc import CPCContainer
 from fastlane_bot.tools.optimizer import MargPOptimizer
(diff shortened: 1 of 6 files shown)


fastlane_bot/modes/base_triangle.py
```diff
@@ -45,16 +46,36 @@ def find_arbitrage(self) -> Dict[List[Any], List[Any]]:
 
         return {"combos": combos, "arb_opps": sorted(arb_opps, key=lambda arb_opp: arb_opp["profit"], reverse=True)}
 
-def get_params(CCm, dst_tokens, src_token):
+def get_params(self, CCm, dst_tokens, src_token):
+    # For a triangle, the pstart of each dst_token is derived from its rate vs the src_token.
+    # Since Carbon orders can contain diverse prices independent of external market prices, and
+    # we require that the pstart be on the Carbon curve to get successful optimizer runs,
+    # then for Carbon orders only we must randomize the pstart from the list of available Carbon curves.
+    # Random selection chosen as opposed to iterating over all possible combinations.
```
barakman, line 54: Why (and since when) do we "require that the pstart be on the Carbon curve to get successful optimizer runs"?

The code below actually opts for a `pstart` which is **NOT** on a Crabon curve, as it literally sorts the list of curves such that all Carbon curves are at the end of that list, and it then picks the price of the first curve on that list.

---

What exactly does choosing a random carbon strategy to get the price from achieve here "as opposed to iterating over all possible combinations"?

First of all, It's not like we were iterating these curves prior to this change.
We simply chose the first curve on the list (`CC[0]`), and then used its price for `pstart`.

Second, even if for some reason we need a random Carbon curve here, the first curve is just as random as any other curve on that list, since you have no knowledge of how this list was generated to begin with.

Third, adding randomness in a system requires a really (REALLY) good reason, as it turns the behavior of that system inconsistent, unexpected, non-deterministic, hard to test and hard to reproduce any reported issues.

For example, reducing the probability of collisions between different bot operators is a good reason for choosing a random arbitrage-opportunity (at the top layer). Such reason does not appear to be viable in the case at hand.

barakman (changes requested):

barakman (changes requested):

PR #677, opened 2024-05-27
fix: get topics from correct event contract abi

platonfloria:


fastlane_bot/events/exchanges/base.py
```diff
@@ -51,6 +52,9 @@ def get_pools(self) -> List[Pool]:
         """
         return list(self.pools.values())
 
+    def get_event_contract(self, w3: Union[Web3, AsyncWeb3]) -> Union[Contract, AsyncContract]:
+        return w3.eth.contract(abi=self.get_abi())
```
barakman, line 56: Can add `.events` at the end of this line, and simply the rest of the code.
Don't forget ro fix the function name and return-value type, if you choose to do it this way.

barakman (changes requested):
I think that we can (and probably should) change the whole scheme to use `contract.events.MyEvent.getLogs`, so we get rid of the `topics` input parameter altogether, and revert back to a method which looks similar to the one previous used (taking only `from_block` and `to_block as input parameters).
Technically, you'll need pass a `contract.events.MyEvent` instance to the `get_events` function, and that function will call function `getLogs` on that instance).

fastlane_bot/events/exchanges/base.py
platonfloria: Purpose of this method is to return a particular contract, not just events. It would be too small of a scope in terms of reusability if it only returned events.

platonfloria:
> I think that we can (and probably should) change the whole scheme to use `contract.events.MyEvent.getLogs`, so we get rid of the `topics` input parameter altogether, and revert back to a method which looks similar to the one previous used (taking only `from_block` and `to_block as input parameters). Technically, you'll need pass a `contract.events.MyEvent`instance to the`get_events`function, and that function will call function`getLogs` on that instance).

Perhaps, but definitely this is outside of the scope of this PR

barakman:
> > I think that we can (and probably should) change the whole scheme to use `contract.events.MyEvent.getLogs`, so we get rid of the `topics` input parameter altogether, and revert back to a method which looks similar to the one previous used (taking only `from_block` and `to_block as input parameters). Technically, you'll need pass a `contract.events.MyEvent`instance to the`get_events`function, and that function will call function`getLogs` on that instance).
> 
> Perhaps, but definitely this is outside of the scope of this PR

In a way it is, because you're already working hard in this PR to create the contracts and their list of events, so you can just replace the `topic` in the `Subscription` class with the actual event instance.

fastlane_bot/events/exchanges/base.py
barakman: Yeah, I'm suggesting to replace this function with a `get_events` function, since we don't seem to need the one currently implemented.

platonfloria:
> > > I think that we can (and probably should) change the whole scheme to use `contract.events.MyEvent.getLogs`, so we get rid of the `topics` input parameter altogether, and revert back to a method which looks similar to the one previous used (taking only `from_block` and `to_block as input parameters). Technically, you'll need pass a `contract.events.MyEvent`instance to the`get_events`function, and that function will call function`getLogs` on that instance).
> > 
> > 
> > Perhaps, but definitely this is outside of the scope of this PR
> 
> In a way it is, because you're already working hard in this PR to create the contracts and their list of events, so you can just replace the `topic` in the `Subscription` class with the actual event instance.

In future I want to try fetching all topics within a single request. Event.get_logs will not work with that.
That is another reason to solve this in a separate PR

barakman (approved):

PR #681, opened 2024-05-27
use max_priority_fee 0 in case of Sei

zavelevsky:


fastlane_bot/config/network.py
```diff
@@ -285,9 +285,12 @@ class ConfigNetwork(ConfigBase):
     # HOOKS
     #######################################################################################
     @staticmethod
-    def gas_strategy(web3):
+    def gas_strategy(web3, avoid_max_priority_fee=False):
         gas_price = web3.eth.gas_price # send `eth_gasPrice` request
-        max_priority_fee = web3.eth.max_priority_fee # send `eth_maxPriorityFeePerGas` request
+        max_priority_fee = 0
+        if not avoid_max_priority_fee:
```
sklbancor, line 291: on the assumption that the request fails quickly, why don't we just do try / except and return fee=0 in case it fails with this particular error? avoids the change in the other file

sklbancor (commented):
so code comment

PR #684, opened 2024-05-30
Add triangle_multi_complete mode

NIXBNT:


fastlane_bot/modes/base_triangle.py
```diff
@@ -42,21 +42,46 @@ def get_triangle_groups(flt, x_y_pairs):
         triangle_groups += [("/".join(sorted([flt,x])), pair, "/".join(sorted([flt,y])))]
     return triangle_groups
 
+def get_all_relevant_pairs_info(CCm, all_relevant_pairs, carbon_v1_forks):
```
zavelevsky, line 45: moving this function up makes it hard to review the actual changes - please return it to the original position for the sake of code diffs

zavelevsky (changes requested):

PR #698, opened 2024-07-25
Add support for Celo blockchain

Lesigh-3100:
Added support & static data for Celo blockchain.

This includes:
- Carbon
- Uniswap 
- Sushiswap
- Ubeswap

The bot skips the wrap/unwrap processor for Celo as the native & wrapped Celo token is treated as a single token.


fastlane_bot/config/network.py
```diff
@@ -821,6 +824,47 @@ class _ConfigNetworkSei(ConfigNetwork):
...
+
+    NATIVE_GAS_TOKEN_ADDRESS = "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"
+    WRAPPED_GAS_TOKEN_ADDRESS = "0x471EcE3750Da237f93B8E339c536989b8978a438"
+    NATIVE_GAS_TOKEN_SYMBOL = "CELO"
+    WRAPPED_GAS_TOKEN_SYMBOL = "WCELO"
+    STABLECOIN_ADDRESS = "0xcebA9300f2b948710d2653dD7B07f33A8B32118C"
+
+    IS_INJECT_POA_MIDDLEWARE = True
+    # No Balancer fork with significant liquidity
+    BALANCER_VAULT_ADDRESS = "0x0000000000000000000000000000000000000000"
```
barakman, line 855: 0xE9265892B5c56264d60e26451862B576814185C9 according to https://github.com/bancorprotocol/bancor-arbitrage/pull/73/files#diff-4b2d495f255090f4d66ca3805482795d6e0e9fead5da674077daf7307e26e9afR224

barakman (changes requested):

fastlane_bot/config/network.py
Lesigh-3100: That address holds $14 total in tokens. I think it would be a mistake to include it as it provides no real value.

